PR #14, opened 2020-05-17
chore: builds linux and mac binaries

greenkiwi:
This makes the following changes:
* It builds binaries for both Linux (linux_amd64) & MacOS (darwin_amd_64)
* The files are places in `out/<platform>` to match the directory structure that TF expects
* It reads in the `VERSION` and appends that to the file names to match TFs expectations

It also adds a `docker-compose.yaml` file that allows for builds inside of docker
* `make docker-go-build` will run the commands in the docker-compose container.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .terraform
 # Provider binary
 terraform-provider-cyral
+out/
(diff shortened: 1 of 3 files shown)


wcmjunior:
All changes proposed here are already in `main`, thus closing this PR. Thanks @greenkiwi !

PR #55, opened 2021-05-19
Changes keycloak service account initialization script

### LINKED ISSUE #54: Script message with unwanted grave quotes
# Description

When running the main defined to be able to login with keycloak based applications, the last `echo` contains backquotes which cause an error on bash, because the commands don't exist.

```bash
echo "Use both `client_id` and `client_secret` returned above to set up your provider."
echo "See the provider documentation in https://github.com/cyralinc/terraform-provider-cyral/blob/main/doc/provider.md how to set those two parameters."
echo
```

ptcar2009:
## Description of the change

Changes service account creation script to prompt for user input and to check for errors at each step. Changes README to suggest using the script instead of having the script in the markdown itself.

The script asks the user for both the CP target and the token, but takes prioritarely from environment variables named `CONTROL_PLANE` and `TOKEN`. The token is to be inserted without the `Authorization: Bearer` because the script already handles that.

Closes #54

## Type of change

### Development

- [x] Lint rules pass locally

### Code review 

### Testing
Created a service account using both the prompted and the env var based approaches, did it successfully. Errors cause errors and are properly printed on the screen.

Did not test the changes to the README because the script is not on the `main` branch, but should work once it is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,11 +31,6 @@ A `Service Account` for the Terraform provider must be created using the [script
 curl https://raw.githubusercontent.com/cyralinc/terraform-provider-cyral/main/scripts/create-service-account-keycloak.sh -O
 bash create-service-account-keycloak.sh
 ```
-or
-
-```bash
-curl https://raw.githubusercontent.com/cyralinc/terraform-provider-cyral/main/scripts/create-service-account-keycloak.sh | bash
-```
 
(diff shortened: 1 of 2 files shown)


wcmjunior (approved):
Nice work. Thanks, @ptcar2009 !

PR #65, opened 2021-06-29
Upgrade terraform plugin SDK to v2.7.0. 

ahmarzouk:
## Description of the change

> Upgrade terraform plugin SDK to v2.7.0. 

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review 


go.mod
```diff
@@ -4,7 +4,5 @@ go 1.15
 
 require (
 	github.com/aws/aws-sdk-go v1.38.45
-	github.com/hashicorp/terraform-plugin-sdk v1.17.2 // indirect
-	github.com/hashicorp/terraform-plugin-sdk/v2 v2.6.1
```
pramodc, line 8: We are importing this sdk(terraform-plugin-sdk) from our source code, right? So we do need this in the go.mod and commit it as well. Whenever we want to update this version, we remove the lines, do go build to update go.mod and commit again.

ahmarzouk: This is intentionally removed to automatically upgrade the plugin SDK to the latest v2.x every time we build the terraform provider which is especially needed for the nightly test.   

removing the dependency from the `go.mod` while still using it in the source code will make go search for the latest version before building the go project as shown below: 
```
go: finding module for package github.com/hashicorp/terraform-plugin-sdk/v2/plugin
go: finding module for package github.com/hashicorp/terraform-plugin-sdk/v2/diag
go: finding module for package github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema
go: found github.com/hashicorp/terraform-plugin-sdk/v2/plugin in github.com/hashicorp/terraform-plugin-sdk/v2 v2.7.0
go: found github.com/hashicorp/terraform-plugin-sdk/v2/diag in github.com/hashicorp/terraform-plugin-sdk/v2 v2.7.0
go: found github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema in github.com/hashicorp/terraform-plugin-sdk/v2 v2.7.0
```

If this doesn't seem like a good solution for this problem, then, @wcmjunior, we might need to manually upgrade the SDK as needed.

wcmjunior: @pramodc , please let us know your thoughts given the explanation above. Also, in the case you agree with the solution, if we should also remove `go.sum` and rebuild both files as part of our `Makefile`.

pramodc: @wcmjunior @ahmarzouk This does not look like how module versioning is typically handled. Go modules is a way to have control over your package version dependencies, this is the reason why both go.mod and go.sum get committed to git in most of the open projects out there (including all of our repos). We want to have complete control over what version of dependency we pick and hence we do this manually.  Without fixing the versions, we may have a scenario where our code compiles one day and doesn't the next because some dependent package broke backward compatibility. Is there a reason why you want to pick the latest version all the time? Have you seen this way of managing in other provider projects?

wcmjunior: @ahmarzouk , just had a call with Pramod and the best way to conduct this from now on is to keep upgrading the SDK as they become obsolete. There is no better way to do this as if we remove the version from `go.mod`, we will never know which version was used to generate a given binary when we cut a release. Anyway, we need to better understand why Hashicorp acts on this way with the SDK as this is quite annoying.

wcmjunior (approved):

PR #111, opened 2021-09-28
Fix omitting access_duration bug in identity map resource

### LINKED ISSUE #81: Omitting access_duration in cyral_identity_map results in plan change on every plan
Link to doc: https://github.com/cyralinc/terraform-provider-cyral/blob/main/doc/resource_identity_map.md

Even though it says that `access_duration` can be omitted, after an `identity_map` has been created, any subsequent terraform plan results in a change. Nothing ends up happening after applying, but these changes show up in every plan thereafter.

```
  ~ resource "cyral_identity_map" "identity_map" {
        id                          = "1v3ApyyPb6VzZGi79c1KSPfZTBL-my-local-account"
        # (4 unchanged attributes hidden)

      - access_duration {
          - days    = 0 -> null
          - hours   = 0 -> null
          - minutes = 0 -> null
          - seconds = 0 -> null
        }
    }
```

I would assume that if the docs specify that its not required, then if its omitted when creating the resource, it shouldn't result in this change on any subsequent plans.

It feels like theres probably some error of not setting this block in the TF state by default if its omitted:
```
      access_duration {
          days    = 0 -> null
          hours   = 0 -> null
          minutes = 0 -> null
          seconds = 0 -> null
       }
```

As a separate, but related question:
its not super clear if setting 
```
    access_duration {
        days    = 0
        hours   = 0
        minutes = 0
        seconds = 0
    }
```
is equivalent to omitting it.

VictorGFM:
## Description of the change

This PR intends to fix an issue reported about omitting the access duration block in the identity map resource. Also, the PR contains some refactorings made in this resource.

Closes #81.

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Development

- [x] Lint rules pass locally

### Code review 

### Testing
Tested running "go test -v -run=^TestAccRepositoryIdentityMapResource$" and through manual terraform configuration file.


cyral/provider.go
```diff
@@ -84,7 +84,6 @@ func Provider() *schema.Provider {
 
 		ResourcesMap: map[string]*schema.Resource{
 			"cyral_datamap":                     resourceDatamap(),
-			"cyral_identity_map":                resourceIdentityMap(),
```
wcmjunior, line 87: Please keep the old registered resource as `deprecated`. We should just remove it in the next major release. You may use the same code for both, but add a parameter to inform that one of the calls are deprecated. See also next suggestions to understand the whole idea.

```suggestion
			"cyral_identity_map":                resourceRepositoryIdentityMap(true),
```

cyral/provider.go
```diff
@@ -102,6 +101,7 @@ func Provider() *schema.Provider {
 			"cyral_repository_conf_analysis":    resourceRepositoryConfAnalysis(),
 			"cyral_repository_conf_auth":        resourceRepositoryConfAuth(),
 			"cyral_repository_local_account":    resourceRepositoryLocalAccount(),
+			"cyral_repository_identity_map":     resourceRepositoryIdentityMap(),
```
wcmjunior, line 104:
```suggestion
			"cyral_repository_identity_map":     resourceRepositoryIdentityMap(false),
```

cyral/resource_cyral_repository_identity_map.go
```diff
@@ -189,14 +194,14 @@ var ReadIdentityMapConfig = ResourceOperationConfig{
 			d.Get("identity_name").(string),
 			d.Get("repository_local_account_id").(string))
 	},
-	ResponseData: &IdentityMapAPIResponse{},
+	ResponseData: &RepositoryIdentityMapAPIResponse{},
 }
 
-func resourceIdentityMap() *schema.Resource {
+func resourceRepositoryIdentityMap() *schema.Resource {
```
wcmjunior, line 200:
```suggestion
func resourceRepositoryIdentityMap(deprecate bool) *schema.Resource {
	deprecationMessage := ""
	if deprecate {
		deprecationMessage = "Use `cyral_repository_identity_map` instead."
	}
```

cyral/resource_cyral_repository_identity_map.go
```diff
@@ -189,14 +194,14 @@ var ReadIdentityMapConfig = ResourceOperationConfig{
...
 	return &schema.Resource{
 		CreateContext: CreateResource(
```
wcmjunior, line 202:
```suggestion
		DeprecationMessage: deprecationMessage,
		CreateContext: CreateResource(
```

wcmjunior (changes requested):
See required changes.

wcmjunior (approved):

PR #125, opened 2021-11-04
Replace sso by idp

wcmjunior:
Replace SSO by IdP for better cohesion and clarity.


VictorGFM (approved):
Great refactoring! I just added a few changes in documentation, but is now good to go!

PR #148, opened 2021-12-16
ENG-7416: Fixes issue about missing sidecar returning 500 error

VictorGFM:
## Description of the change

This PR fixes an issue with the sidecar resource about returning a `500` error when the sidecar was actually not found. So the fix was basically handling the error and just removing the resource from the state. Notice that, currently, the sidecar API always returns a status code of `500` for every error, so its not possible to distinguish if the error returned is related to a `404 Not Found` or not by its status code. This way, a workaround for that is to check if the error message matches a 'Failed to extract info for wrapper' message, since thats the current message returned when the sidecar is not found. Once this issue is fixed in the sidecar API, we should handle the error here, on the Terraform Provider side, by its status code, and only remove the resource from the state (d.SetId("")) if it returns a 404 Not Found.

## Type of change

### Development

- [x] Lint rules pass locally

### Code review 

### Testing
Manual tests through Terraform config, and also running all the acceptance tests on `Edge` CP by typing `make` after setting the env vars.

### DIFF
--- a/cyral/resource_cyral_sidecar.go
+++ b/cyral/resource_cyral_sidecar.go
@@ -112,7 +112,7 @@ func resourceSidecarRead(ctx context.Context, d *schema.ResourceData, m interfac
 			log.Printf("[DEBUG] Sidecar not found. SidecarID: %s. "+
 				"Removing it from state. Error: %v", d.Id(), err)
 			d.SetId("")
-			return diag.Diagnostics{}
+			return nil
 		}
 
 		return createError(fmt.Sprintf("Unable to read sidecar. SidecarID: %s",
--- a/cyral/resource_cyral_sidecar.go
+++ b/cyral/resource_cyral_sidecar.go
@@ -97,13 +97,21 @@ func resourceSidecarRead(ctx context.Context, d *schema.ResourceData, m interfac
 
 	body, err := c.DoRequest(url, http.MethodGet, nil)
 	if err != nil {
-		return createError(fmt.Sprintf("Unable to read sidecar. SidecarID: %s",
-			d.Id()), fmt.Sprintf("%v", err))
+		// Currently, the sidecar API always returns a status code of 500 for every error,
+		// so its not possible to distinguish if the error returned is related to
+		// a 404 Not Found or not. This way, once this is fixed in the sidecar API,
+		// we should handle the error here by its status code, and only remove the
+		// resource from the state (d.SetId("")) if it returns a 404 Not Found,
+		// otherwise we should return the error in the Diagnostics.
+		log.Printf("[DEBUG] Unable to read sidecar. SidecarID: %s. "+
+			"Removing it from state. Error: %v", d.Id(), err)
+		d.SetId("")
+		return diag.Diagnostics{}
 	}
 
 	response := SidecarData{}
 	if err := json.Unmarshal(body, &response); err != nil {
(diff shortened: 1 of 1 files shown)


VictorGFM:
I noticed that none of our resources currently handles if the read response was a `404 Not Found`, so that it can remove the resource from the state to allow the Terraform to recognize that and update the Plan correctly. So that probably mean that we have to update all the resources to handle this type of scenario. Please see:

- [Repository Resource not handling 404 Not Found](https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/resource_cyral_repository.go#L95)
- [Datamap Resource not handling 404 Not Found](https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/resource_cyral_datamap.go#L107)

cyral/resource_cyral_sidecar.go
```diff
@@ -97,13 +97,21 @@ func resourceSidecarRead(ctx context.Context, d *schema.ResourceData, m interfac
...
-			d.Id()), fmt.Sprintf("%v", err))
+		// Currently, the sidecar API always returns a status code of 500 for every error,
+		// so its not possible to distinguish if the error returned is related to
+		// a 404 Not Found or not. This way, once this is fixed in the sidecar API,
+		// we should handle the error here by its status code, and only remove the
+		// resource from the state (d.SetId("")) if it returns a 404 Not Found,
+		// otherwise we should return the error in the Diagnostics.
+		log.Printf("[DEBUG] Unable to read sidecar. SidecarID: %s. "+
+			"Removing it from state. Error: %v", d.Id(), err)
+		d.SetId("")
```
ptcar2009, line 108: If the configuration for the control plane is incorrect, I imagine there will also be an error there, as well as with connection issues and etc. I don't think just checking if there was an error is enough, we need to know if the error is actually a "does not exist" error before removing it from the state, otherwise there will be side effects in the future.

ptcar2009 (changes requested):

cyral/resource_cyral_sidecar.go
VictorGFM: That makes sense, although we are not able to check the not found error by its status code, so the only workaround solution I see is to check if the error message matches the `Failed to
extract info for wrapper` message, which is returned when the sidecar is not found. Otherwise, we will have to fix it in the sidecar API.

wcmjunior: @VictorGFM , I related this ticket internally in Jira to the one that will fix the API.

cyral/resource_cyral_sidecar.go
```diff
@@ -97,13 +98,30 @@ func resourceSidecarRead(ctx context.Context, d *schema.ResourceData, m interfac
...
+		// issue is fixed in the sidecar API, we should handle the error here by its status
+		// code, and only remove the resource from the state (d.SetId("")) if it returns a 404
+		// Not Found.
+		matched, regexpError := regexp.MatchString("Failed to extract info for wrapper",
+			err.Error())
+		if regexpError == nil && matched {
+			log.Printf("[DEBUG] Sidecar not found. SidecarID: %s. "+
+				"Removing it from state. Error: %v", d.Id(), err)
+			d.SetId("")
+			return diag.Diagnostics{}
```
ptcar2009, line 115: It's better to explicitly return `nil` here.
```suggestion
			return nil
```

ptcar2009 (approved):

PR #152, opened 2021-12-28
ENG-7170: Terraform provider creates a new identity map instead of updating old one when state changes

ptcar2009:
## Description of the change

- Changes identity map resource to delete previous resource and create new one on update
- CHanges id for identity map resource to be independent of fields

## Type of change

### Development

- [x] Lint rules pass locally

### Code review

### Testing
Created, updated, deleted identity maps with valid and invalid parameters, all worked as intended.

### DIFF
--- a/cyral/resource_cyral_repository_identity_map.go
+++ b/cyral/resource_cyral_repository_identity_map.go
@@ -271,6 +271,7 @@ func resourceRepositoryIdentityMap(deprecate bool) *schema.Resource {
 			"identity_name": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 			"access_duration": {
 				Type:     schema.TypeSet,
--- a/cyral/resource_cyral_repository_identity_map.go
+++ b/cyral/resource_cyral_repository_identity_map.go
@@ -256,14 +256,17 @@ func resourceRepositoryIdentityMap(deprecate bool) *schema.Resource {
 			"repository_id": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 			"repository_local_account_id": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 			"identity_type": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 			"identity_name": {
 				Type:     schema.TypeString,


cyral/resource_cyral_repository_identity_map.go
```diff
@@ -218,37 +232,63 @@ func resourceRepositoryIdentityMap(deprecate bool) *schema.Resource {
...
+			return ReadResource(readConfig)(ctx, d, m)
+		},
+
 		DeleteContext: DeleteResource(
 			ResourceOperationConfig{
 				Name:       "RepositoryIdentityMapResourceDelete",
 				HttpMethod: http.MethodDelete,
 				CreateURL: func(d *schema.ResourceData, c *client.Client) string {
-					return fmt.Sprintf("https://%s/v1/repos/%s/identityMaps/%s/%s/%s",
+					return fmt.Sprintf("https://%s/v1/repos/%s",
```
wcmjunior, line 289: I don't think it is a good idea to define the URL from the ID as these are different concepts. It also makes the code confusing and one may think that the URL is `https://CP_ADDRESS/v1/repos/SOME_ID` and just realize it the actual endpoint is different if the resource id is actually analyzed.

cyral/resource_cyral_repository_identity_map.go
```diff
@@ -218,37 +232,63 @@ func resourceRepositoryIdentityMap(deprecate bool) *schema.Resource {
...
+			if _, err := c.DoRequest(createNewUrl, http.MethodPut, resourceData); err != nil {
+				return createError("Unable to create new integration", fmt.Sprintf("%v", err))
+			}
+
+			d.SetId(fmt.Sprintf("%s/identityMaps/%s/%s/%s",
+				d.Get("repository_id").(string),
+				d.Get("identity_type").(string),
+				d.Get("identity_name").(string),
+				d.Get("repository_local_account_id").(string)),
+			)
```
wcmjunior, line 270: 1. This will break backwards compatibility with previous resource. If this is the only way to fix the issue, then I'm ok with that.
2. So far, we used `-` as a separator for those IDs that are composed of several elements. We should use the same set of practices in the whole provider unless it is not possible;
3. Using the ID to compose the URL makes the code harder to read as you need to check how the ID is created in order to guess what is the resource endpoint.

```suggestion
			d.SetId(fmt.Sprintf("%s-%s-%s-%s",
				d.Get("repository_id").(string),
				d.Get("identity_type").(string),
				d.Get("identity_name").(string),
				d.Get("repository_local_account_id").(string)),
			)
```

cyral/resource.go
```diff
@@ -74,11 +75,22 @@ func ReadResource(config ResourceOperationConfig) schema.ReadContextFunc {
 
 		body, err := c.DoRequest(url, config.HttpMethod, nil)
 		if err != nil {
+
+			if config.UnsetIdOnError {
+				d.SetId("")
+				return nil
+			}
```
wcmjunior, line 82: Could you please elaborate on how this will prevent the issue described in the PR title? I thought the only change required is how we handle the ID.

wcmjunior:
@VictorGFM , would like to hear your thoughts on this PR as well.

VictorGFM:
> @VictorGFM , would like to hear your thoughts on this PR as well.

Sure, I'll take a look.

VictorGFM (changes requested):

cyral/resource_cyral_repository_identity_map.go
ptcar2009: If the url is not defined from the ID everything breaks because of what we've discussed over time. There's no way to not attach those two, since the url depends on the id, and the url is unique in the same way as the id.

cyral/resource_cyral_repository_identity_map.go
ptcar2009: This just makes we need to have splits and joins all over the place. The `/`s are for the URL. This is a pattern among other providers such as gcloud and other things, and most of our IDs are actual IDs, and that's why we use `-`, but in this case it's different, and having `-` just makes things more complicated, in my eye.

cyral/resource.go
ptcar2009: If we don't do this, everything breakes once something is deleted upstream, and it's a simple change, backwards compatible that allows for lot's of little future improvements in the behavior for our resources.

cyral/resource.go
```diff
@@ -57,7 +58,7 @@ func CreateResource(createConfig ResourceOperationConfig, readConfig ResourceOpe
 		}
 		log.Printf("[DEBUG] Response body (unmarshalled): %#v", createConfig.ResponseData)
 
-		createConfig.ResponseData.WriteToSchema(d)
+		createConfig.ResourceData.WriteToSchema(d)
```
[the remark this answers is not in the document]
ptcar2009: The return `ReadResource` already uses the `ResponseData` to write to the schema, which makes this call redundant.

cyral/resource_cyral_repository_identity_map.go
```diff
@@ -218,37 +232,63 @@ func resourceRepositoryIdentityMap(deprecate bool) *schema.Resource {
...
+
+			removeUrl := fmt.Sprintf("https://%s/v1/repos/%s",
+				c.ControlPlane,
+				d.Id(),
+			)
+
+			if _, err := c.DoRequest(removeUrl, http.MethodDelete, resourceData); err != nil {
```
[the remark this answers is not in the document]
ptcar2009: You might have thought of mentioning this before hahaha. But we need some changes to the id logic anyway.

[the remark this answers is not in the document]
wcmjunior: @ptcar2009 , move to `ForceNew` instead.

cyral/resource.go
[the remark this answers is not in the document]
VictorGFM: Not really, depending on the struct type that is being passed and the different implementations of the `WriteToSchema` method, this could end up changing the resource behavior. For instance, consider the `resource_cyral_integration_datadog`, it calls [this WriteToSchema](https://github.com/cyralinc/terraform-provider-cyral/blob/8d97b8ee2c48cde216be794d405c95903d7711d3/cyral/resource.go#L33) in the `CreateResource`, but in the `ReadResource` a [different WriteToSchema](https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/resource_cyral_integration_datadog.go#L17) is called, this is also valid for other resources. Btw, I agree that the `resource.go` needs a general refactoring, but ideally, this needs to be done in a separate and specific task, since a lot of resources depend on this file.

ptcar2009:
I've removed everything I did and just added the `ForceNew` clauses. Works as intended with backwards compatibility.

cyral/resource_cyral_repository_identity_map.go
```diff
@@ -256,14 +256,17 @@ func resourceRepositoryIdentityMap(deprecate bool) *schema.Resource {
...
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 			"identity_type": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 			"identity_name": {
```
VictorGFM, line 271: I tried updating the `identity_name`, but this created a new resource without deleting the old one, so this argument should include the `ForceNew` too. For some reason, this is only not necessary for the access_duration. Other than that, everything else looks great, just add this change and I'll approve it.

VictorGFM (changes requested):

VictorGFM (approved):
LGTM!

PR #190, opened 2022-04-21
ENG-8268: Add support for certificate bundle secrets in sidecar resource

wcmjunior:
## Description of the change

 Add new parameter `certificate_bundle_secrets` to `cyral_sidecar`.

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review

### Testing

Unit tests and manual testing in `edge`.


cyral/resource_cyral_sidecar.go
```diff
@@ -184,11 +240,90 @@ func getSidecarDataFromResource(c *client.Client, d *schema.ResourceData) (Sidec
...
+	ret := make(CertificateBundleSecrets)
+
+	if len(rdCBS) > 0 {
+		cbsMap := rdCBS[0].(map[string]interface{})
+		for k, v := range cbsMap {
+			vList := v.(*schema.Set).List()
+			// 1. k = "sidecar" or other direct internal elements of certificate_bundle_secrets
+			// 2. Also one element on this list due to MaxItems...
+			// 3. Ignore self signed certificates
+			if len(vList) > 0 && k != "sidecar-generated-selfsigned" {
```
VictorGFM, line 304: Notice that `k = "sidecar"` will always be true since the schema only supports certificate bundle secret for the `sidecar` key, so there's no need to check `k` value here.

VictorGFM (commented):
I've tested the resource and overall it seems to be working just fine! I have a few comments about the code though.

cyral/resource_cyral_sidecar.go
wcmjunior: Not always. See comment in `1`. What I'm trying to do here is not mess with the `sidecar-generated-selfsigned` element as it does not matter for the provider, but all the others, including `sidecar` does. In this sense, I don't want the provider to break when we add a new type of `CertificateBundleSecret` with a different name. So far, only `sidecar-generated-selfsigned` and `sidecar` exist, but we need to be resilient to other types in the future.

VictorGFM:
Had to update the Makefile to fix an issue in the test script introduced by [this PR](https://github.com/cyralinc/terraform-provider-cyral/pull/204).

VictorGFM (approved):
LGTM, re-tested everything and ran all the automated tests, working just fine.

PR #193, opened 2022-05-02
ENG-8392: Change repo local account resource to support Kubernetes Secret

VictorGFM:
## Description of the change

ENG-8392: Terraform provider - add resource for k8s secrets local account

- Add support for Kubernetes Secret
- Deprecate `enviroment_variable` field since it contains a typo
- Fix issue with `repository_id` by adding ForceNew to this field
- Fix some details in docs.

## Type of change

### Development

- [x] Lint rules pass locally

### Code review

### Testing

Manual and automated tests

### DIFF
--- a/cyral/resource_cyral_repository_local_account.go
+++ b/cyral/resource_cyral_repository_local_account.go
@@ -37,13 +37,13 @@ func (resource *AwsIAMResource) ReadFromSchema(d *schema.ResourceData) {
 	}
 }
 
-type AwsSecretsResource struct {
+type AwsSecretsManagerResource struct {
 	DatabaseName string `json:"databaseName"`
 	RepoAccount  string `json:"repoAccount"`
 	SecretArn    string `json:"secretARN"`
 }
 
-func (resource AwsSecretsResource) WriteToSchema(d *schema.ResourceData) {
+func (resource AwsSecretsManagerResource) WriteToSchema(d *schema.ResourceData) {
 	d.Set("aws_secrets_manager", []interface{}{
 		map[string]interface{}{
 			"database_name": resource.DatabaseName,
@@ -53,7 +53,7 @@ func (resource AwsSecretsResource) WriteToSchema(d *schema.ResourceData) {
 	})
 }
 
-func (resource *AwsSecretsResource) ReadFromSchema(d *schema.ResourceData) {
+func (resource *AwsSecretsManagerResource) ReadFromSchema(d *schema.ResourceData) {
 	data := d.Get("aws_secrets_manager").(*schema.Set)
 
 	for _, id := range data.List() {
@@ -121,14 +121,14 @@ func (resource *HashicorpVaultResource) ReadFromSchema(d *schema.ResourceData) {
 	}
 }
 
-type EnviromentVariableResource struct {
+type EnvironmentVariableResource struct {
 	DatabaseName string `json:"databaseName"`
 	RepoAccount  string `json:"repoAccount"`
 	VariableName string `json:"variableName"`
 }
 
-func (resource EnviromentVariableResource) WriteToSchema(d *schema.ResourceData) {
-	d.Set("enviroment_variable", []interface{}{
+func (resource EnvironmentVariableResource) WriteToSchema(d *schema.ResourceData) {
+	d.Set("environment_variable", []interface{}{
 		map[string]interface{}{
 			"database_name": resource.DatabaseName,
 			"local_account": resource.RepoAccount,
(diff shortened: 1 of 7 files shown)


VictorGFM:
@wcmjunior , while I was working on these changes I noticed some limitations regarding the way the provider is structured/organized, more specifically about the way the [resource.go](https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/resource.go) is currently designed. This file defines a [SchemaDataHandler](https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/resource.go#L16) interface with two methods: [WriteToSchema](https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/resource.go#L17) and [ReadFromSchema](https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/resource.go#L18), they are used to manage the CRUD operations for almost all the resources. What I noticed is that currently if an error occurs within these methods, is not possible to return and handle these errors which limit the way we can manage the resource lifecycles. In this case, I intended to add a [Config](https://github.com/cyralinc/repo-auth-client-go/blob/main/messages/repo_auth.proto#L269) field to the [repository_local_account](https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/resource_cyral_repository_local_account.go) resource, since this field is present in our API but is missing in the resource, but because of this limitation I mentioned, there was not a good way to add the `Config` field to the `repository_local_account` (since it involved handling a parse operation that could throw an error) and I also ended up spending a considerable time figuring a way to solve that.

That being said, I also believe that there are other problems with the `resource.go` regarding readability and other aspects, because of that , even though this is not something of high priority, I think we should consider refactoring this file, which would require us to create a specific task for that, since this file is used by almost all the resources we have. Let me know what you think.

cyral/resource_cyral_repository_local_account.go
```diff
@@ -121,31 +121,86 @@ func (resource *HashicorpVaultResource) ReadFromSchema(d *schema.ResourceData) {
...
+		})
+	} else {
+		d.Set("environment_variable", []interface{}{
+			map[string]interface{}{
+				"database_name": resource.DatabaseName,
+				"local_account": resource.RepoAccount,
+				"variable_name": resource.VariableName,
+			},
+		})
+	}
```
wcmjunior, line 149: Had a hard time finding the actual difference and saw that it was the typo in `environment`. I think the best make this code easier to follow is to just have in the conditional the two strings with and without the typo as follows:
```suggestion
	argument := "environment_variable"
	if useDeprecatedArgument {
		// Deprecated: should be removed in the next MAJOR release
		argument = "enviroment_variable"
	}
	d.Set(argument, []interface{}{
		map[string]interface{}{
			"database_name": resource.DatabaseName,
			"local_account": resource.RepoAccount,
			"variable_name": resource.VariableName,
		},
	})
```

cyral/resource_cyral_repository_local_account.go
```diff
@@ -269,15 +322,11 @@ func resourceRepositoryLocalAccount() *schema.Resource {
...
-		ConflictsWith: []string{
-			"aws_iam",
-			"cyral_storage",
-			"hashicorp_vault",
-			"enviroment_variable",
-		},
+	awsSecretsManagerSchema := &schema.Schema{
+		Type:         schema.TypeSet,
+		Optional:     true,
+		ExactlyOneOf: secretManagersTypes,
```
wcmjunior, line 328: Much better than using the `ConflictsWith`. :-)

cyral/resource_cyral_repository_local_account.go
```diff
@@ -378,14 +440,44 @@ func resourceRepositoryLocalAccount() *schema.Resource {
 		},
 	}
 
+	kubernetesSecretSchema := &schema.Schema{
```
wcmjunior, line 443:
```suggestion
	kubernetesSecretSchema := &schema.Schema{
		Description: "",
```

cyral/resource_cyral_repository_local_account.go
```diff
@@ -378,14 +440,44 @@ func resourceRepositoryLocalAccount() *schema.Resource {
...
+		Type:         schema.TypeSet,
+		Optional:     true,
+		ExactlyOneOf: secretManagersTypes,
+		MaxItems:     1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"database_name": {
```
wcmjunior, line 450:
```suggestion
				"database_name": {
					Description: "",
```

wcmjunior (commented):
Please add descriptions to the schema as in [this PR](https://github.com/cyralinc/terraform-provider-cyral/pull/183).

cyral/resource_cyral_repository_local_account.go
VictorGFM: That's way better, thanks for the suggestion!

wcmjunior (approved):

ccampo133 (approved):

VictorGFM:
> Sorry, @VictorGFM , I did not see this message. TBH, we didn't spend much though on this `resource.go`. The goal back then was just to eliminate code replication as we had several resources that would look mostly the same. I'm totally in favor of refactoring it, but we should spend more time discussing the pros and cons of any new approach. Please create an issue in this repo so we can let this one for public discussion. Thanks!

Sure, I'll create an issue for that.

PR #239, opened 2022-07-01
Add data source for cyral_integration_idp

wcmjunior:
## Description of the change

Add a new data source `cyral_integration_idp` to list configured identity providers.

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review

### Testing

Performed manual tests with individual filters, all filters and no filters. Unit tests will be added in a next patch.


cyral/data_source_cyral_integration_idp.go
```diff
@@ -0,0 +1,145 @@
...
+	if idpIntegrations.Connections != nil {
+		for _, connection := range idpIntegrations.Connections.Connections {
+			log.Printf("[DEBUG] Connection: %#v", connection)
+			if connection != nil {
+				// Conditions to return data:
+				// 1. displayName is not empty and correspond to the one in the current connection;
+				// 2. idpType is not empty and correspond to the first characters of the current connection alias;
+				// 3. displayName and idpType are empty.
+				if (displayName != "" && displayName == connection.DisplayName) ||
+					(idpType != "" && idpType == connection.Alias[0:len(idpType)]) ||
```
VictorGFM, line 126: Instead of using `connection.Alias[0:len(idpType)]` may be better to use `strings.HasPrefix`, since it also prevents `slice out of bounds` errors

cyral/data_source_cyral_integration_idp.go
```diff
@@ -0,0 +1,145 @@
...
+				Description: "Filter results by the IdP integration type.",
+				Optional:    true,
+				Type:        schema.TypeString,
+				ValidateFunc: validation.StringInSlice([]string{
+					"aad",
+					"adfs",
+					"forgerock",
+					"gsuite",
+					"okta",
+					"ping_one",
```
VictorGFM, line 52: The alias of Ping One integrations are prefixed by `pingone` instead, changing this should be enough to fix filtering by this type of IdP.
```suggestion
					"pingone",
```

VictorGFM (changes requested):
Did some tests with the data source, there seem to have a few points that we could change:

cyral/data_source_cyral_integration_idp.go
```diff
@@ -0,0 +1,145 @@
...
+				}, false),
+			},
+			"idp_set": {
+				Type: schema.TypeSet,
```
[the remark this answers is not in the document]
wcmjunior: Good point!

cyral/data_source_cyral_integration_idp.go
```diff
@@ -0,0 +1,167 @@
...
+						},
+						"enabled": {
+							Description: "True if the IdP integration is enabled.",
+							Type:        schema.TypeBool,
+							Computed:    true,
+						},
+					},
+				},
+				Computed:    true,
+				Description: "Set of existing IdP integrations for the given filter criteria.",
```
VictorGFM, line 84:
```suggestion
				Description: "List of existing IdP integrations for the given filter criteria.",
```

cyral/data_source_cyral_integration_idp.go
```diff
@@ -0,0 +1,167 @@
...
+
+	log.Printf("[DEBUG] display_name: %s", displayNameFilter)
+	log.Printf("[DEBUG] type: %s", idpTypeFilter)
+	if idpIntegrations.Connections != nil {
+		for _, connection := range idpIntegrations.Connections.Connections {
+			log.Printf("[DEBUG] Connection: %#v", connection)
+			if connection != nil {
+				// Skip in case filters are non-empty but
+				if (displayNameFilter != "" && displayNameFilter != connection.DisplayName) ||
+					(idpTypeFilter != "" && !strings.HasPrefix(connection.Alias, idpTypeFilter)) {
```
VictorGFM, line 131: Since the API is already filtering by type, this line is redundant
```suggestion
					(idpTypeFilter != "" && !strings.HasPrefix(connection.Alias, idpTypeFilter)) {
```

cyral/data_source_cyral_integration_idp.go
```diff
@@ -0,0 +1,167 @@
...
+				// 2. displayNameFilter and idpTypeFilter are empty.
+				// 3. idpTypeFilter is not empty (already filtered in the endpoint call)
+				// Conditions NOT to return data:
+				// 4. displayNameFilter is not empty and does not correspond to the one in the current connection OR
+				//    idpTypeFilter is not empty and does not correspond to the one in the current connection
+				if ((displayNameFilter != "" && displayNameFilter == connection.DisplayName) || // 1
+					(displayNameFilter == "" && idpTypeFilter == "") || // 2
+					(idpTypeFilter != "")) && // 3
+					!((displayNameFilter != "" && displayNameFilter != connection.DisplayName) || // 4
+						(idpTypeFilter != "" && !strings.HasPrefix(connection.Alias, idpTypeFilter))) {
```
VictorGFM, line 145: Why do we need these conditionals? The API will already filter by type, then we just need to worry if it needs to be filtered by display_name, but we already check that in the previous `if(displayNameFilter != "" && displayNameFilter != connection.DisplayName)`.

If we reach this part of the code, the idp is already filtered by type (in the API call) and by `display_name`, since it passed the `if` mentioned above. Also, if no filter is passed, then the API will not filter by type and the ìf`above will not filter by `display_name` also. I think we can just remove these conditionals and just append the item to the `idpList`
```suggestion
```

VictorGFM (approved):
Added two comments, other than that the data source looks good!

cyral/data_source_cyral_integration_idp.go
wcmjunior: Indeed. I removed the unused code. Thanks.

PR #242, opened 2022-07-07
ENG-8736: Add data source to retrieve repositories

Yowgf:
## Description of the change

This PR adds the data source `cyral_repository`. This data source provides access to existing repositories in the Cyral control plane. One is able to filter the retrieved repositories by a regex on their name, or by their type (e.g. `sqlserver`).

## Type of change

### Development

- [x] Lint rules pass locally

### Code review

### Testing

ACC tests are passing.

Manually deploying the following works: 

```terraform

resource "cyral_repository" "sqlserver1" {
    host = "localhost"
    port = 1433
    type = "sqlserver"
    name = "prefix-sqlserver1"
}

resource "cyral_repository" "mongodb1" {
    host = "localhost"
    port = 27017
    type = "mongodb"
    name = "prefix-mongodb1"
    properties {
        mongodb_replica_set {
            max_nodes = 2
            replica_set_id = "replica-set-id"
        }
    }
}

data "cyral_repository" "test_repository" {
    depends_on = [
      cyral_repository.sqlserver1,
      cyral_repository.mongodb1,
    ]
    name = "^prefix-m.*1$"
    # type = "mongodb"
}

output "repodata" {
  value = jsonencode(data.cyral_repository.test_repository.repository_list)
}
```

In this case, the `repodata` output only contains information about the MongoDB repo.


cyral/resource_cyral_repository.go
```diff
@@ -16,8 +16,41 @@ import (
 const (
 	mongodbRepoType             = "mongodb"
 	mongodbReplicaSetServerType = "replicaset"
+	repositoryTypesMarkdown     = "\n  - `bigquery`" +
```
wcmjunior, line 19: This should be created out of `repositoryTypes` to avoid type duplication.

cyral/resource_cyral_repository.go
```diff
@@ -16,8 +16,41 @@ import (
...
+		"dremio",
+		"galera",
+		"mariadb",
+		"mongodb",
+		"mysql",
+		"oracle",
+		"postgresql",
+		"redshift",
+		"snowflake",
+		"s3",
```
wcmjunior, line 49:
```suggestion
		"s3",
		"snowflake",
```

Put `s3` before `snowflake` so we can have the same order in `repositoryTypesMarkdown`.

VictorGFM (changes requested):
Hey @Yowgf, the overall code looks good! Although I found an issue when testing the following config on `edge.dev`:
```
data "cyral_repository" "test_repository" {
    type = "sqlserver"
}
```
It gives the following error:
```
Stack trace from the terraform-provider-cyral_v2.7.0 plugin:

panic: set item just set doesn't exist
```
I tried to debug and see if I could find the reason for it but I didn't figure it out. It seems to be related to the repo `properties` field. Could you take a look and see if you can fix that?

cyral/resource_cyral_repository.go
Yowgf: Thanks Wilson!  Fixed.

cyral/resource_cyral_repository.go
```diff
@@ -18,6 +18,33 @@ const (
...
+	}
+}
+
+func repositoryTypesMarkdown() string {
+	var s string
+	for _, repoType := range repositoryTypes() {
+		s += fmt.Sprintf("\n  - `%s`", repoType)
+	}
+	return s
+}
```
wcmjunior, line 46: Same code in Scala just to make you cry with me:

```scala
def repositoryTypesMarkdown = repositoryTypes.map(repoType => s"\n - `$repoType`")
```

:-P

Yowgf:
Thanks  @VictorGFM, thanks for pointing that out! I just fixed the bug. The issue happened because some repositories on `edge` had non-null `properties` field, and were not replica-sets. In that case, we would add an empty map to the Terraform argument `properties`. When doing `d.Set("properties", malformedObject)`, a panic in the Terraform code is triggered.

VictorGFM (approved):
Looks good @Yowgf, thanks for adding this :)

Issue #268, opened 2022-08-08
Issues with the properties argument of the repository resource


VictorGFM:
1) The `properties` argument of the `cyral_repository` resource is generating a non-empty plan at every `terraform apply` when it is specified as an empty block. One way to fix that would be to validate if the `properties` block is empty, and if it is we should require that at least one property is specified (`AtLeastOneOf` can be used for that).

Config:
```terraform
resource "cyral_repository" "pg_repo" {
  name = "pgrepo"
  type = "postgresql"
  host = "some-host"
  port = 5432
  properties {}
}
```

Result:
```terraform
Terraform will perform the following actions:

  # cyral_repository.pg_repo will be updated in-place
  ~ resource "cyral_repository" "pg_repo" {
        id     = "2D5gZ7EsdCwmutxcO9VNSUrXcpX"
        name   = "pgrepo"
        # (4 unchanged attributes hidden)

      + properties {
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

2) Also, there's another issue regarding the `properties` argument on the version `2.7.1` of the provider. If a data repository object has a non-empty `properties` field, but it does not match the following condition:
```go
data.Properties != nil && data.Properties.MongoDBServerType == mongodbReplicaSetServerType
```
An error occurs since the provider will try to set an empty map to the `properties` argument (which is a `TypeSet`).
This issue was already fixed in version `2.8.0`, but we should create a patch to fix it for `2.7.x`.

Repo object example:
```json
{
    "repo": {
        "name": "pgrepo",
        "type": "postgresql",
        "repoHost": "some-host",
        "repoPort": 5432,
        "connParams": {
            "connDraining": {
                "auto": false,
                "waitTime": 0
            }
        },
        "labels": [],
        "properties": {
            "test": "test"
        }
    },
    "sidecars": null
}
```

Error logs:
```
panic: set item just set doesn't exist

goroutine 103 [running]:
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*MapFieldWriter).setSet(0xc00068fe30, {0xc00090ef50, 0x1, 0xc00007a3c0}, {0x17a37e0, 0xc00068fec0}, 0xc000420dc0)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.17.0/helper/schema/field_writer_map.go:327 +0x972
[...]

PR #271, opened 2022-08-12
ENG-9133: Fix issues with repository resource properties

Yowgf:
## Description of the change

Fixes issue [268](https://github.com/cyralinc/terraform-provider-cyral/issues/268).

## Type of change

### Development

- [x] Lint rules pass locally

### Code review

### Testing

Manually test the configs presented in the issue. Also, add an ACC test that captures the issue. The TestAccRepositoryResource test is passing.

### DIFF
--- a/cyral/resource_cyral_repository.go
+++ b/cyral/resource_cyral_repository.go
@@ -40,12 +40,13 @@ func (data *RepoData) WriteToSchema(d *schema.ResourceData) {
 	d.Set("name", data.Name)
 	d.Set("labels", data.Labels)
 
-	properties := data.PropertiesAsInterface()
-	d.Set("properties", properties)
+	if properties := data.PropertiesAsInterface(); properties != nil {
+		d.Set("properties", properties)
+	}
 }
 
 func (data *RepoData) PropertiesAsInterface() []interface{} {
-	properties := []interface{}{}
+	var properties []interface{}
 	if data.Properties != nil {
 		if data.IsReplicaSet() {
 			propertiesMap := make(map[string]interface{})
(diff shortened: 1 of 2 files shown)


cyral/resource_cyral_repository.go
```diff
@@ -33,35 +33,38 @@ type RepoData struct {
...
 
-	var properties []interface{}
-	if data.Properties != nil {
-		propertiesMap := make(map[string]interface{})
+	properties := data.PropertiesAsInterface()
+	d.Set("properties", properties)
+}
 
+func (data *RepoData) PropertiesAsInterface() []interface{} {
+	properties := []interface{}{}
```
VictorGFM, line 48: I think this is more idiomatic and it's also simpler than the `DiffSuppressFunc` logic. Please change to that instead and remove the `DiffSuppressFunc`, thanks.
```suggestion
	if properties := data.PropertiesAsInterface(); properties != nil {
		d.Set("properties", properties)
	}
}

func (data *RepoData) PropertiesAsInterface() []interface{} {
	var properties []interface{}
```

cyral/resource_cyral_repository.go
```diff
@@ -152,6 +155,16 @@ func resourceRepository() *schema.Resource {
...
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+					propertiesSet := d.Get("properties").(*schema.Set)
+					if k == "properties.#" && oldValue == "0" && newValue == "1" && propertiesSet.Len() == 1 {
+						mongodbConfig := propertiesSet.List()[0]
+						if mongodbConfig == nil {
+							return true
+						}
+					}
+					return false
+				},
```
VictorGFM, line 167: No need to use `DiffSuppressFunc`, please see my comment above.
```suggestion
```

VictorGFM (changes requested):
Thanks for fixing this @Yowgf, please see my comment below on how to make this fix simpler.

cyral/resource_cyral_repository.go
Yowgf: This works! Thanks Victor!

I don't know why I didn't try this. Let me at least explain why I went through so much trouble, maybe it will help you in the future too, to not get confused as I did. If you type in `properties {}` or NOTHING in your Terraform config, you will get the same value from `d.Get("properties")` -- `[]interface{}{}` in go syntax. I was trying to distinguish between the two configs with this, so it didn't work.

Also, there was another really strange issue when you had a non-mongodb repo: the error message only showed up on the first `terraform apply`. If you entered `terraform apply` again, you would be able to create a replica set sqlserver repo, for example. However, just moving the line 

```go
		properties = new(RepositoryProperties)
```

down into the `for` seemed to have fixed the issue (see latest commit).

Anyway, I just committed the changes. Thanks!

VictorGFM (approved):
LGTM @Yowgf!

PR #280, opened 2022-09-13
ENG-9386: add resource AccessRules

ricardorey10:
## Description of the change

This PR adds a resource for the Gatekeeper's Access Rules.

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review

### Testing
Did a lot of manual testing running `terraform apply`, `terraform import`, etc, and things look sane AFAICT. Acceptance tests are also passing.


cyral/resource_cyral_access_rules.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+		m["validFrom"] = rule.ValidFrom
+		m["validUntil"] = rule.ValidUntil
+		m["config"] = map[string]interface{}{
+			"authorizationPolicyInstanceIDs": rule.Config.AuthorizationPolicyInstanceIDs,
+		}
+
+		rules = append(rules, m)
+	}
+	d.Set("rule", rules)
```
Yowgf, line 59:
```suggestion
	return d.Set("rule", rules)
```

`d.Set` returns an error. Although we don't expect this to err, and we currently ignore this error in a lot of places in our code, I think it is best if we propagate it.

cyral/resource_cyral_access_rules.go
```diff
@@ -0,0 +1,292 @@
...
+		Description: "Manage access rules",
+		CreateContext: CreateResource(
+			ResourceOperationConfig{
+				Name:       "RepositoryAccessRulesCreate",
+				HttpMethod: http.MethodPost,
+				CreateURL: func(d *schema.ResourceData, c *client.Client) string {
+					d.SetId(
+						marshalComposedID(
+							[]string{
+								d.Get("repository_id").(string),
```
Yowgf, line 140: Maybe we can declare a variable like `repositoryID` before `d.SetId` and reuse it here and when creating the URL. Same for `user_account_id`.

cyral/resource_cyral_access_rules.go
```diff
@@ -0,0 +1,292 @@
...
+		),
+		DeleteContext: DeleteResource(
+			ResourceOperationConfig{
+				Name:       "RepositoryAccessRulesDelete",
+				HttpMethod: http.MethodDelete,
+				CreateURL: func(d *schema.ResourceData, c *client.Client) string {
+
+					idPieces, err := unmarshalComposedID(d.Id(), "/", 2)
+					if err != nil {
+						panic("Failed to unmarshal access rules ID")
```
Yowgf, line 190:
```suggestion
						panic("Failed to unmarshal access rules ID: " + err.Error())
```
I know we don't expect this to panic, but maybe it would be nice to include the error information, just in case.

cyral/resource_cyral_access_rules.go
```diff
@@ -0,0 +1,292 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Description: "ID of this resource in Cyral environment.",
+				Computed:    true,
+				Type:        schema.TypeString,
+			},
+
+			"repository_id": {
+				Description: "ID of the repository",
```
Yowgf, line 212:
```suggestion
				Description: "ID of the repository.",
```
Just a minor detail (the dot).

FYI, the description is copied into the markdown that goes to the Provider docs, so you can write markdown syntax here if you wish :).

Yowgf (commented):
The resource looks good, thanks Ricardo! I added some comments regarding some really minor details.

Also, when you finish the resource, you will want to generate the docs:

```
make docker-compose/docs
pre-commit run --show-diff-on-failure --color=always --all-files
```

In my machine, `make docker-compose/docs` is kind of bugged, it runs as a different user, so I need to run something like  `find docs -exec chown myuser {} \;` after it. If you could fix that it would be very much appreciated :).

cyral/resource_cyral_access_rules.go
ricardorey10: Done. Thanks for the suggestion!

cyral/resource_cyral_access_rules.go
ricardorey10: Done.

cyral/resource_cyral_access_rules.go
ricardorey10: Good idea. Done.

cyral/resource_cyral_access_rules.go
ricardorey10: Yeah, thanks for catching this, I was pretty inconsistent with punctuation. Done.

And thanks for the suggestion about markdown!

cyral/provider.go
```diff
@@ -171,6 +171,7 @@ func Provider() *schema.Provider {
 			"cyral_role_sso_groups":             resourceRoleSSOGroups(),
 			"cyral_sidecar":                     resourceSidecar(),
 			"cyral_sidecar_credentials":         resourceSidecarCredentials(),
+			"cyral_repository_access_rules":     resourceRepositoryAccessRules(),
```
VictorGFM, line 177: Maybe its a good idea to have this defined near the group of the other `cyral_repository*` resources

ricardorey10:
@Yowgf @VictorGFM @wcmjunior I still need to add unit tests, but I think the "core" logic should be good for review now.

Please let me know if I should also look into the "data_source" stuff, or if the resource is good enough for now.

cyral/resource_cyral_access_rules.go
```diff
@@ -0,0 +1,329 @@
...
+	return &schema.Resource{
+		Description: "Manage access rules",
+		CreateContext: CreateResource(
+			ResourceOperationConfig{
+				Name:       "RepositoryAccessRulesCreate",
+				HttpMethod: http.MethodPost,
+				CreateURL: func(d *schema.ResourceData, c *client.Client) string {
+					repoID := d.Get("repository_id").(string)
+					userAccountID := d.Get("user_account_id").(string)
+					d.SetId(
```
VictorGFM, line 140: The resource ID is usually set after the request succeeds on creating the resource. This can be done in the WriteToSchema of the `AccessRulesResponse`.

cyral/resource_cyral_access_rules.go
```diff
@@ -0,0 +1,329 @@
...
+			},
+			ReadRepositoryAccessRulesConfig,
+		),
+		DeleteContext: DeleteResource(
+			ResourceOperationConfig{
+				Name:       "RepositoryAccessRulesDelete",
+				HttpMethod: http.MethodDelete,
+				CreateURL: func(d *schema.ResourceData, c *client.Client) string {
+
+					idPieces, err := unmarshalComposedID(d.Id(), "/", 2)
```
VictorGFM, line 191: Instead of having to unmarshal the composed ID, maybe we could use the d.Get("repository_id").(string) and d.Get("user_account_id").(string) instead. I think this would simplify the code, but the way it is also works just fine.

cyral/resource_cyral_access_rules.go
```diff
@@ -0,0 +1,329 @@
...
+						repoID,
+						userAccountID,
+					)
+				},
+			},
+		),
+
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Description: "ID of this resource in Cyral environment.",
```
VictorGFM, line 209: I think would be good to describe how the ID is composed (E.g.: {repository_id}/{user_account_id}), or we could just include a note on the documentation similar to what's done for the identity map resource (https://github.com/cyralinc/terraform-provider-cyral/blame/main/docs/resources/repository_identity_map.md#L5)). This would help users to understand how they should be importing the resource.

cyral/provider.go
VictorGFM: We should also rename the file name to `resource_cyral_repository_access_rules.go` to match the resource name

cyral/resource_cyral_access_rules.go
```diff
@@ -0,0 +1,329 @@
...
+						},
+
+						"valid_until": {
+							Description: "The end time for the grant. Format is: " +
+								"`yyyy-mm-ddThh:mm:ssZ`. Eg. `2022-01-24T18:30:00Z`.",
+							Optional: true,
+							Type:     schema.TypeString,
+						},
+
+						"config": {
```
VictorGFM, line 277: Would be good to set `MaxItems: 1` to the `config`.

VictorGFM (commented):
At a high level looks very good @ricardorey10. I provided some initial comments that I think could help.

docs/resources/repository_access_rules.md
```diff
@@ -0,0 +1,60 @@
+---
```
ricardorey10, line 1: This file still needs some work. In particular, I need to create a template file with instructions on how the id is calculated, like we do for https://github.com/cyralinc/terraform-provider-cyral/blob/main/docs/resources/repository_identity_map.md. My understanding is that this file is somehow built from a template. I'll have to look into this.

ricardorey10:
@wcmjunior @VictorGFM @Yowgf Acceptance tests have not yet been written for this resource, because access rules live under user accounts, and we don't have the TF resource for user accounts yet.

Hannah was going to work on that, but she's out sick now, so I'll take that task in order to get this ready ASAP. Thus, I'll pause this PR for a while, finish the one for user accounts, and *then* get back to this one, once we have the ability to create user accounts via Terraform.

wcmjunior:
Let's merge this changes only after we release `v2.9.0` (please reach out to me if you need more info) to avoid maintaining two production branches at this time. We will very likely have to keep a `v2` branch in the future, but I prefer to wait to avoid incoming cherry-picks. FYI @VictorGFM .

cyral/provider.go
ricardorey10: Done both.

cyral/resource_cyral_access_rules.go
ricardorey10: The problem here is that the creation API does not return an ID, or any of the pieces we need to compute it. The only place where we have the full information is here (that's why Alex suggested I could set the ID here, IIRC).

cyral/resource_cyral_access_rules.go
ricardorey10: I feel like this is slightly cleaner, as we use the marshal function to compute the ID and then the unmarshal function to read it, so if it's OK then I'll just keep this as-is

cyral/resource_cyral_access_rules.go
ricardorey10: Yeah, I think having this in the docs precisely how we do it for that stuff you linked will be the cleanest way to do this.

cyral/resource_cyral_access_rules.go
ricardorey10: Did not think about that, but makes sense. Done.

wcmjunior:
@ricardorey10 , please don 't forget to change the target to `main` instead.

cyral/resource_cyral_access_rules.go
ricardorey10: Ah, I keep forgetting what things we have to extract from the API call and which ones we get "for free". I'll fix this, thanks for the suggestion.

ricardorey10:
@VictorGFM @Yowgf Quick poke about reviewing this :grin:

cyral/resource_cyral_repository_access_rules.go
```diff
@@ -0,0 +1,335 @@
...
+										Required:    true,
+										Type:        schema.TypeString,
+									},
+								},
+							},
+							MinItems: 1,
+							MaxItems: 1,
+						},
+
+						"valid_from": {
```
VictorGFM, line 268: We might want to make `valid_from` a required field depending on the conclusion of this [discussion](https://cyralinc.slack.com/archives/C0173L983EU/p1664240598935429).

VictorGFM (approved):
The changes look good Ricardo! I provided minor comments below that may require a few changes. Im already approving the PR though.

examples/resources/cyral_repository_access_rules/resource.tf
```diff
@@ -0,0 +1,36 @@
+resource "cyral_repository_access_rules" "some_resource_name" {
```
Yowgf, line 1: I am facing an infinite `terraform apply` loop when trying out this example, where even though my .tf file hasn't changed, the Terraform plan kept suggesting this change:

```
Terraform will perform the following actions:

  # cyral_repository_access_rules.t1 will be updated in-place
  ~ resource "cyral_repository_access_rules" "t1" {
        # (2 unchanged attributes hidden)

      ~ rule {
            # (2 unchanged attributes hidden)

          + config {}

            # (1 unchanged block hidden)
        }
      ~ rule {
          + config {}

            # (1 unchanged block hidden)
        }

        # (1 unchanged block hidden)
    }
```

Don't know how to fix it out of the top of my head though :/

Yowgf (commented):
Looks good, Ricardo! Just some comments to improve the tests.

cyral/resource_cyral_repository_access_rules_test.go
```diff
@@ -0,0 +1,410 @@
...
+					ImportState:       true,
+					ImportStateVerify: true,
+					ImportStateIdFunc: importStateComposedIDFunc(
+						importStateResName,
+						[]string{
+							"repository_id",
+							"user_account_id",
+						},
+						"/",
+					),
```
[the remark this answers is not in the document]
ricardorey10: For my own understanding: what is that used for, and why is it not needed for this test?

cyral/resource_cyral_repository_access_rules_test.go
```diff
@@ -0,0 +1,410 @@
...
+					ResourceName: importStateResName,
+				},
+			},
+		},
+	)
+}
+
+func setupRepositoryAccessRulesTest(
+	integrationData *AccessRulesResource,
```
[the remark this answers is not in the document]
ricardorey10: Done.

cyral/resource_cyral_repository_access_rules_test.go
```diff
@@ -0,0 +1,410 @@
...
+	bareBones bool,
+) (string, resource.TestCheckFunc) {
+	var configuration string
+	configuration += formatBasicRepositoryIntoConfig(
+		basicRepositoryResName,
+		"tfprov-test-repository-access-rules-repository",
```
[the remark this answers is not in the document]
ricardorey10: Sure thing, I can do this.

cyral/resource_cyral_repository_access_rules_test.go
```diff
@@ -0,0 +1,410 @@
...
+		"mongodb",
+		"mongo.local",
+		3333,
+	) + "\n"
+
+	configuration += userAccConfig(basicRepositoryID, "user_acc_for_tf_access_rules") + "\n"
```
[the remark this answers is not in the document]
ricardorey10: Done.

examples/resources/cyral_repository_access_rules/resource.tf
ricardorey10: Can you share with me the contents of your `.tf` file?

cyral/resource_cyral_repository_access_rules_test.go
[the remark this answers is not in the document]
Yowgf: When running an import test, the framework grabs the id of the resource referenced by the field ResourceName, and sends it over as a parameter to the importer.

ImportStatIdFunc is used to generate this id in a custom manner. It was used in the past in some resources where the actual ID format and import ID syntax were not the same, so we needed to grab for example the repository_id from the resource state and set that to be the id before passing it to the importer.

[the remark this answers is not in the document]
ricardorey10: Gotcha. Thanks for the explanation. I thought I had deleted this code, as you suggested, but it looks like I actually forgot about it. Done with the deletion now.

examples/resources/cyral_repository_access_rules/resource.tf
Yowgf: I have pasted it below. It is essentially the configuration found in `examples/resources/cyral_repository_access_rules/resource.tf`.

```terraform
resource "cyral_repository" "t1" {
  name = "test1"
  type  = "sqlserver"
  port = 1433
  host = "sqlserver.local"
}

resource "cyral_repository_user_account" "t1" {
  repository_id = cyral_repository.t1.id
  name = "my-user-account"
  auth_scheme {
    aws_secrets_manager {
      secret_arn = "redacted"
    }
  }
}

resource "cyral_repository_access_rules" "t1" {
  repository_id = cyral_repository.t1.id
  user_account_id = cyral_repository_user_account.t1.id

  # First rule to be considered
  rule {
    identity {
      type = "username"
      name = "my-user-account"
    }
    config {
      policy_ids = []
    }
    valid_from = "2022-09-19T23:50:48.606Z"
    valid_until = "2022-10-10T00:00:00Z"
  }

  # Second rule to be considered
  rule {
    identity {
      type = "email"
      name = "fake-username@gmail.com"
    }
  }

  # ... and so on
  rule {
    identity {
      type = "group"
      name = "fake-group"
    }
    config {
      policy_ids = []
    }
  }
}
```

ricardorey10:
@VictorGFM @Yowgf there is a mess going on with the branches in this PR. I have no idea why, but even though I was on top of Hannah's PR the base branch suddenly changed, and it's really messy.

Once Hannah merges her user accounts stuff I'll rebase on top of main and make sure that everything is fine.

PR #281, opened 2022-09-14
ENG-9398: New resource: Listeners API

gengdahlCyral:
## Description of the change

Adding terraform provider for new listeners CRUD API.
See Jira: https://cyralinc.atlassian.net/browse/ENG-9398

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review

### Testing

> Description of testing


cyral/rescource_cyral_listener.go
```diff
@@ -0,0 +1,177 @@
...
+// POST {{baseURL}}/sidecars/:sidecarID/listeners/ (Create a listener)
+// PUT {{baseURL}}/sidecars/:sidecarID/listeners/:listenerID (Update a listener)
+// DELETE {{baseURL}}/sidecars/:sidecarID/listeners/:listenerID (Delete a listener)
+func resourceSidecarListener() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Manages [sidecar listeners](https://cyral.com/docs/sidecars/sidecar-listeners).",
+		CreateContext: resourceSidecarListenerCreate,
+		ReadContext:   resourceSidecarListenerRead,
+		UpdateContext: resourceSidecarListenerUpdate,
+		DeleteContext: resourceSidecarListenerDelete,
```
ricardorey10, line 60: I believe it's more "idiomatic" to use `CreateResource()`, `ReadResource()`, `UpdateResource()`, and `DeleteResoure()`, (see https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/resource_cyral_repository_identity_map.go#L211) instead of creating ad-hoc functions.

This will also lead to simpler code, because you just need to implement `ReadFromSchema()` for create/update, and `WriteToSchema()` for read.

gengdahlCyral: Thanks @ricardorey10 for putting me on the right path, rewriting this.

Yowgf (commented):
This looks very good @gengdahlCyral ! I added some comments below that I hope you will find helpful.

cyral/resource_cyral_listener.go
```diff
@@ -0,0 +1,276 @@
...
+				Computed:    true,
+			},
+			"sidecar_id": {
+				Description: "ID of the sidecar that the listener will be bound to.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+			},
+			"repo_types": {
+				Description: "List of repository types that the listener supports. Currently limited to one repo type, eg [\"mysql\"]",
```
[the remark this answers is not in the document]
Yowgf: Actually, I thought about this a bit more, it seems best to leave it as is. I feel like we will forget to remove the `MaxItems: 1` when more repo types become supported.

cyral/resource_cyral_listener.go
```diff
@@ -0,0 +1,276 @@
...
+
+type ReadSidecarListenersAPIResponse struct {
+	Listener *SidecarListener `json:"listener"`
+}
+type CreateListenerAPIResponse struct {
+	ListenerId string `json:"listenerId"`
+}
+
+func (c CreateListenerAPIResponse) WriteToSchema(d *schema.ResourceData) error {
+	d.SetId(c.ListenerId)
```
Yowgf, line 65:
```suggestion
	d.SetId(marshalComposedID([]string{d.Get("sidecar_id"), c.ListenerId}, "/"))
```

It is best to keep a consistency between the ID the user uses to import the resource in:

```
terraform import {sidecar_id}/{listener_id}
```

and the actual ID of the Terraform resource. I believe the average Terraform user would expect this consistency.

`terraform import -h` displays the following message:

```
Usage: terraform [global options] import [options] ADDR ID

...

  The ADDR specified is the address to import the resource to. Please
  see the documentation online for resource addresses. The ID is a
  resource-specific ID to identify that resource being imported. Please
  reference the documentation for the resource type you're importing to
  determine the ID syntax to use. It typically matches directly to the ID
  that the provider uses.
```

cyral/resource_cyral_listener.go
```diff
@@ -0,0 +1,276 @@
...
+				Description: "File in which the sidecar will listen for the given repository. Required for unix listeners and mutual exclusive with tcp_listener_port.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"tcp_listener_port": {
+				Description: "Port in which the sidecar will listen for the given repository. Required for tcp listeners and mutual exclusive with unix_listener_file.",
+				Type:        schema.TypeInt,
+				Optional:    true,
+			},
+			"tcp_listener_host": {
```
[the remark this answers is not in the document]
gengdahlCyral: yes, you listen using tcp or unix file. 
I'm letting API do all validations. There are plenty of them [here](https://github.com/cyralinc/crud/blob/main/wrapper/listeners/listeners_validation.go)

gengdahlCyral (commented):
I think I've addressed all your comments @Yowgf

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,415 @@
...
+				Computed:    true,
+			},
+			SidecarIdKey: {
+				Description: "ID of the sidecar that the listener will be bound to.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+			},
+			RepoTypesKey: {
+				Description: "List of repository types that the listener supports. Currently limited to one repo type, eg [\"mysql\"]",
```
VictorGFM, line 296:
```suggestion
				Description: "List of repository types that the listener supports. Currently limited to one repo type, eg [\"mysql\"].",
```

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,415 @@
...
+					},
+				},
+			},
+			MultiplexedKey: {
+				Description: "Multiplexed listener, defaults to not multiplexing (false). Not supported for all repository types.",
+				Type:        schema.TypeBool,
+				Optional:    true,
+			},
+			TcpListenerKey: {
+				Description: "tcp listener settings.",
```
VictorGFM, line 325:
```suggestion
				Description: "TCP listener settings.",
```

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,415 @@
...
+						PortKey: {
+							Description: "Port in which the sidecar will listen for the given repository.",
+							Type:        schema.TypeInt,
+							Required:    true,
+						},
+					},
+				},
+			},
+			MysqlSettingsKey: {
+				Description: "MysqlSettings represents the listener settings for a mysql data repository.",
```
VictorGFM, line 346: I think since this is a user-facing description we could list the repo types that each particular setting is supposed to support. For instance, for `MysqlSettingsKey` it also supports `mariadb`, so would be good to list it in the description so that this can be more intuitive to users.

VictorGFM (commented):
Did an overall review and looks good @gengdahlCyral, its definitely on the right path. Let me know once you add the tests and think the PR should be good to go and I'll be glad to review it again.

cyral/resource_cyral_sidecar_listener.go
gengdahlCyral: fixed

cyral/resource_cyral_sidecar_listener.go
gengdahlCyral: fixed

cyral/resource_cyral_listener.go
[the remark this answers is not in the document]
Yowgf: I see. I am fine with letting the API handle all the validation. But I believe in terms of UX I believe having as many validations as possible enforced `statically` (before any infrastructure is actually created in `terraform apply`) feels a bit better. If possible, I believe we should add a `ConflictsWith: []string{UnixListenerKey, TcpListenerKey}` to the arguments `"tcp_listener"` and `"unix_listener"`.

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,422 @@
...
+	if len(anInterface) == 0 {
+		return
+	}
+	if anInterface != nil {
+		l.TcpListener = &TcpListener{
+			Host: anInterface[0].(map[string]interface{})[HostKey].(string),
+			Port: anInterface[0].(map[string]interface{})[PortKey].(int),
+		}
+	}
+	log.Printf("[DEBUG] LOLOLO: %s %d", l.TcpListener.Host, l.TcpListener.Port)
```
ricardorey10, line 139: This should go away

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,422 @@
...
+		result[0] = map[string]interface{}{
+			HostKey: l.TcpListener.Host,
+			PortKey: l.TcpListener.Port,
+		}
+		return result
+	}
+	return nil
+}
+func (l *SidecarListener) TcpSettingsFromInterface(anInterface []interface{}) {
+	if len(anInterface) == 0 {
```
ricardorey10, line 130: Maybe

```
if len(anInterface) == 0 || anInterface == nil {
  return
}
```

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,422 @@
...
+		}
+	}
+	log.Printf("[DEBUG] LOLOLO: %s %d", l.TcpListener.Host, l.TcpListener.Port)
+}
+func (l *SidecarListener) UnixSettingsAsInterface() []interface{} {
+	if l.UnixListener != nil {
+		result := make([]interface{}, 1)
+		result[0] = map[string]interface{}{
+			FileKey: l.UnixListener.File,
+		}
```
ricardorey10, line 146: Same as above: this slice can be declared "inline" without having to first `make()` it.

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,422 @@
...
+		return result
+	}
+	return nil
+}
+func (l *SidecarListener) UnixSettingsFromInterface(anInterface []interface{}) {
+	if len(anInterface) == 0 {
+		return
+	}
+	if anInterface != nil {
```
ricardorey10, line 155: As above, these 2 can be combined.

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,422 @@
...
+				Computed:    true,
+			},
+			SidecarIdKey: {
+				Description: "ID of the sidecar that the listener will be bound to.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+			},
+			RepoTypesKey: {
+				Description: "List of repository types that the listener supports. Currently limited to one repo type, eg [\"mysql\"].",
```
ricardorey10, line 303: Maybe you can use https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/utils.go#L67 like it is being used here: https://github.com/cyralinc/terraform-provider-cyral/blob/main/cyral/resource_cyral_repository.go#L118

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,422 @@
...
+						FileKey: {
+							Description: "File in which the sidecar will listen for the given repository.",
+							Type:        schema.TypeString,
+							Required:    true,
+						},
+					},
+				},
+			},
+			MultiplexedKey: {
+				Description: "Multiplexed listener, defaults to not multiplexing (false). Not supported for all repository types.",
```
ricardorey10, line 327: Maybe a note such as "... for all repository types. See "repo_types" for details" would be good

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,422 @@
...
+						PortKey: {
+							Description: "Port in which the sidecar will listen for the given repository.",
+							Type:        schema.TypeInt,
+							Required:    true,
+						},
+					},
+				},
+			},
+			MysqlSettingsKey: {
+				Description: "MysqlSettings represents the listener settings for a [mysql, galera, mariadb] data repository.",
```
ricardorey10, line 353: This is typically spelled as `MySQL`. Since this is a user-facing doc, we should probably stick to that.

Same for the names in the square brackets (they are proper nouns, so we should keep the names as they are typically spelled).

ricardorey10 (commented):
You'll definitely want @VictorGFM and @Yowgf to take a look at this, but I'm leaving some comments of my own :grin:

cyral/resource_cyral_sidecar_listener.go
gengdahlCyral: nice

cyral/resource_cyral_sidecar_listener.go
gengdahlCyral: The values in brackets here refer to the repo type, which is all-lower. Not sure if we want to mix those? 
Anyways, for the rest, you are suggesting 
`MysqlSettingsKey    = "mysql_settings"` -> `MysqlSettingsKey    = "MySQL_settings" `
and the decription to reflect that, ie `"MySQL_settings represents ..."`
or am I misreading it?

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,428 @@
...
+			map[string]interface{}{
+				FileKey: l.UnixListener.File,
+			},
+		}
+		return result
+	}
+	return nil
+}
+func (l *SidecarListener) UnixSettingsFromInterface(anInterface []interface{}) {
+	if len(anInterface) == 0 || anInterface == nil {
```
VictorGFM, line 149:
```suggestion
	if len(anInterface) == 0 {
```

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,428 @@
...
+func (l *SidecarListener) MysqlSettingsFromInterface(anInterface []interface{}) {
+	if len(anInterface) == 0 {
+		return
+	}
+	if anInterface != nil {
+		l.MysqlSettings = &MysqlSettings{
+			DbVersion:    anInterface[0].(map[string]interface{})[DbVersionKey].(string),
+			CharacterSet: anInterface[0].(map[string]interface{})[CharacterSetKey].(string),
+		}
+	}
```
VictorGFM, line 174: We can simplify this by using the following instead: (Also applies to the other fields)
```suggestion
	if len(anInterface) == 0 {
		return
	}
        l.MysqlSettings = &MysqlSettings{
	        DbVersion:    anInterface[0].(map[string]interface{})[DbVersionKey].(string),
	        CharacterSet: anInterface[0].(map[string]interface{})[CharacterSetKey].(string),
        }
```

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,428 @@
...
+func (data ReadSidecarListenersAPIResponse) WriteToSchema(d *schema.ResourceData) error {
+	if data.ListenerConfig != nil {
+		_ = d.Set(ListenerIdKey, data.ListenerConfig.ListenerId)
+		_ = d.Set(RepoTypesKey, data.ListenerConfig.RepoTypesAsInterface())
+		_ = d.Set(UnixListenerKey, data.ListenerConfig.UnixSettingsAsInterface())
+		_ = d.Set(TcpListenerKey, data.ListenerConfig.TcpSettingsAsInterface())
+		_ = d.Set(MultiplexedKey, data.ListenerConfig.Multiplexed)
+		_ = d.Set(S3SettingsKey, data.ListenerConfig.S3SettingsAsInterface())
+		_ = d.Set(DynamoDbSettingsKey, data.ListenerConfig.DynamoDbSettingsAsInterface())
+	}
```
VictorGFM, line 96: This is missing the `MysqlSettings`.

cyral/resource_cyral_sidecar_listener_test.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+func TestSidecarListenerResource(t *testing.T) {
+
+	testConfig, testFunc := setupSidecarListenerTest()
+	resource.ParallelTest(t, resource.TestCase{
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testConfig,
```
VictorGFM, line 64: I would recommend creating different steps for each listener configuration instead of testing everything together in a single step. This is just a suggestion though, feel free to ignore it.

docs/resources/sidecar_listener.md
```diff
@@ -0,0 +1,121 @@
+# cyral_sidecar_listener (Resource)
```
VictorGFM, line 1: Could you please run the `docker-compose/docs` script to update the doc with the latest changes? Thanks

VictorGFM (commented):
Hi @gengdahlCyral, I notice that the `MysqlSettings` is not being set in the `WriteToSchema` during the read operation. Other than that I also provided some minor comments below. Let me know once you fix that and I'll approve the PR, thanks.

cyral/resource_cyral_sidecar_listener.go
gengdahlCyral: updated according to comments from Victor, marking as resolved.

gengdahlCyral:
Hi @VictorGFM Thanks for your review, I've addressed most of them.
Let's hold merge a bit until ENG-9909 gets merged (has listener API payload changes, so better pull them in here).
Also, I've pointed this to feature/multiplexing branch instead (ENG-9909 changes are targeting this branch, so makes sense).

I'll ping you when this PR has been updated to accommodate changes from ENG-9909.

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,363 @@
...
+// create a constant block for schema keys
+
+const (
+	SidecarIdKey        = "sidecar_id"
+	ListenerIdKey       = "listener_id"
+	RepoTypesKey        = "repo_types"
+	NetworkAddressKey   = "network_address"
+	PortKey             = "port"
+	HostKey             = "host"
+	FileKey             = "file"
```
VictorGFM, line 22: I believe the `UnixListener` type got remove from the `ListenerConfig` design, right? If thats case, we may not need this `file` argument anymore.

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,363 @@
...
+				Required:    true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			NetworkAddressKey: {
+				Description:  "The network address that the sidecar listens on.",
+				Type:         schema.TypeSet,
+				Optional:     true, // This field is required, due to ExactlyOneOf directive
+				ExactlyOneOf: []string{NetworkAddressKey},
```
VictorGFM, line 274: Since now we only have the `NetworkAddress` we can just make this field `required` and remove the `ExactlyOneOf` clause.

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,363 @@
...
+						},
+						PortKey: {
+							Description: "Port where the sidecar will listen for the given repository.",
+							Type:        schema.TypeInt,
+							Required:    true,
+						},
+					},
+				},
+			},
+			MySQLSettingsKey: {
```
VictorGFM, line 291: I believe the API already validates that at least one of MySQLSettingsKey, S3SettingsKey or DynamoDbSettingsKey needs to be specified, but another option would be to make this a static validation in the terraform side by changing these arguments to `schema.TypeList` and use the `ExactlyOneOf`. Im usually in favor of static validation (it also helps reflect the correct validations to the auto-generated docs), but this is just a suggestion, feel free to ignore.

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,363 @@
...
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						DbVersionKey: {
+							Description: "MySQL DB version. Required (and only relevant) for listeners of type `mysql`.",
+							Type:        schema.TypeString,
+							Optional:    true,
+						},
+						CharacterSetKey: {
+							Description: "MySQL character set. Optional and only relevant for listeners of type `mysql`.",
```
VictorGFM, line 305: Nit: Just to keep consistency in the docs.
```suggestion
							Description: "MySQL character set. Optional (and only relevant) for listeners of type `mysql`.",
```

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,363 @@
...
+				// Notice the MaxItems: 1 here. This ensures that the user can only specify one this block.
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						ProxyModeKey: {
+							Description: "DynamoDB proxy mode. Only relevant for listeners of type `dynamodb`. Note " +
+								"that `proxy_mode` must be set to `true` for listeners of type `dynamodb`. Defaults to false.",
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
```
VictorGFM, line 342: Not sure if I actually followed this description. If the ProxyMode must be set to true, then why do we need this argument? Seems to me that would be better to completely remove the DynamoDbSettings as theres no much to be configured in this case. @gengdahlCyral why is this field added to the Listener API? I might be missing something here

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,363 @@
...
+	CreateURL: func(d *schema.ResourceData, c *client.Client) string {
+		return fmt.Sprintf("https://%s/v1/sidecars/%s/listeners/%s",
+			c.ControlPlane,
+			d.Get(SidecarIdKey).(string),
+			d.Get(ListenerIdKey).(string))
+	},
+	NewResponseData: func(_ *schema.ResourceData) ResponseData { return &ReadSidecarListenersAPIResponse{} },
+}
+
+type ReadSidecarListenersAPIResponse struct {
```
VictorGFM, line 68:
```suggestion
type ReadSidecarListenerAPIResponse struct {
```

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -0,0 +1,363 @@
...
+}
+func (l *SidecarListener) RepoTypesFromInterface(anInterface []interface{}) {
+	repoTypes := make([]string, len(anInterface))
+	for i, v := range anInterface {
+		repoTypes[i] = v.(string)
+	}
+	l.RepoTypes = repoTypes
+}
+func (l *SidecarListener) NetworkAddressAsInterface() []interface{} {
+	if l.NetworkAddress != nil {
```
VictorGFM, line 109: Nit: We could make the nil checks for all the *AsInterface methods more consistent by using either:
```suggestion
	if l.NetworkAddress == nil {
		return nil
	}
```
or
```suggestion
	if l.NetworkAddress != nil {
		...
	}
	return nil
```

VictorGFM (commented):
@hbaackmann Overall the resource looks good to me, I just have a few comments below that could help improve it. Thanks a lot for adding all those tests, looks very good.

VictorGFM:
Hey @wcmjunior, @gengdahlCyral mentioned that this PR is targeting the `feature/multiplexing` branch since all the other related PRs are being merged to a feature branch and this is not released yet. Does it make sense to do the same in the terraform provider? Im asking this because Im usually not a big fan of feature branches, I was considering merging this PR into main and omitting the listener resource from the ResourcesMap until the API changes get released. What do you think?

cyral/resource_cyral_sidecar_listener.go
hbaackmann: True! Good catch, will remove.

cyral/resource_cyral_sidecar_listener.go
gengdahlCyral: What does this mean? Most of these are optional (unless listener is of type dynamo). 
mysq doesn't need one, nor does s3, or any of the others (oracle, sqlserver, pg etc)

cyral/resource_cyral_sidecar_listener.go
juniocezar: Hi @VictorGFM 

> If the ProxyMode must be set to true, then why do we need this argument? 

Because it dictates how the sidecar will handle that specific TCP connection.


> Seems to me that would be better to completely remove the DynamoDbSettings as there's no much to be configured in this case.

I initially disagree, because even though this field must always be `true` today for DynamoDB, we may allow it to be false in the future and it means if we remove it now and re-add it later, that would bring an unnecessary change in the listener payload, especially because when this field is missing we would not now if we should consider the value to be false or true.


> @gengdahlCyral could you help give the context of why this field was added to the Listener API? I might be missing something here

It was added because the sidecar **supports** both proxy and non-proxy mode for DynamoDB and DynamoDB Streams. However, we are not allowing the configuration with `proxy == false` to be used in an initial moment.

I don't know if this is possible, but if we could have the default as `true` for DynamoDB that would be better, but I don't know if that is feasible here or if that is a good choice in the terraform context.

cyral/resource_cyral_sidecar_listener.go
VictorGFM: @gengdahlCyral you're right, the correct would be to use ConflictsWith instead of ExactlyOneOf, this way the settings would still be optional but wouldn't be able to be specified together for the same listener. [See](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Schema).

cyral/resource_cyral_sidecar_listener.go
VictorGFM: @juniocezar I see, thanks for the context. Given that we might allow `ProxyMode` for dynamoDB to be set to `false` in the future we might want to keep this field then. Regarding the default value, I personally think that the resource should match the API behavior, so in this case - since the API requires the `ProxyMode` to be `true` - I would suggest making the dynamoDB `ProxyMode` a required field (`Required: true`) and keep the default value to `false`, as it is today in the API.

hbaackmann: @VictorGFM I am hesitant to make ProxyMode to required since it is not *strictly* required by the API defintion (see [here](https://github.com/cyralinc/wrapper-client-go/blob/137c6aed3247ca86e4d2988f5ac181d26f23c6af/messages/listeners.proto#L672)-- the validation is enforced in the go code, which is subject to change. However, I think I will go ahead and do it, since it is technically required by the API.

VictorGFM: @hbaackmann yea, I understand your point. From my perspective, keeping the ProxyMode optional for dynamoDB would be inconsistent with the current API behavior, given that today the Listener API requires the ProxyMode to be true for dynamoDB - even though its not explicitly defined in the proto design - and I'm usually in favor that the resource replicates the API behavior (If in the future we make this field optional in the API we should update the resource to match that). That said, I just wanted to share my thought about this here but I think its ok if we decide to keep the ProxyMode optional (even though I think its a bit inconsistent).

hbaackmann:
> Hey @wcmjunior, @gengdahlCyral mentioned that this PR is targeting the `feature/multiplexing` branch since all the other related PRs are being merged to a feature branch while this is not released yet. Does it make sense to do the same in the terraform provider? Im asking this because Im usually not a big fan of feature branches, I was considering merging this PR into `main` and omitting the listener resource from the [ResourcesMap](https://github.com/cyralinc/terraform-provider-cyral/pull/281/files#diff-a86c00a3b0901ce990ddd38a99fa0644e7651c9fc44c091ae09ec1bf7a186147R121) until the API changes get released. What do you think?

@VictorGFM That approach sounds fine to me! Should I go ahead and do that?

VictorGFM (approved):
LGTM, thanks @hbaackmann. I just sent a message to @wcmjunior to confirm which branch this PR should be merged, lets wait to see if we all agree and then we can target `main`.

wcmjunior:
Hey. Sorry for missing the previous messages. :)
Merging to `main` is a safer option, IMO. Disable it from the resources map and make sure the resource will not show up in the docs as well. I just need to know who will be responsible for re-adding the resource to the resources map and then testing everything e2e. We can take this conversation internally.

wcmjunior (commented):
Disable the resource, the tests and the docs. Then we will merge and re-enable as soon as the feature is ready.

PR #283, opened 2022-09-15
ENG-9385: add UserAccount resource

ricardorey10:
## Description of the change

> Description here

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review

### Testing

> Description of testing

### DIFF
--- a/cyral/provider.go
+++ b/cyral/provider.go
@@ -174,6 +174,7 @@ func Provider() *schema.Provider {
 			"cyral_role_sso_groups":             resourceRoleSSOGroups(),
 			"cyral_sidecar":                     resourceSidecar(),
 			"cyral_sidecar_credentials":         resourceSidecarCredentials(),
+			"cyral_provider_database_account":   resourceRepositoryDatabaseAccount(),
 		},
 		ConfigureContextFunc: providerConfigure,
 	}
--- a/cyral/resource_cyral_repository_user_account.go
+++ b/cyral/resource_cyral_repository_user_account.go
@@ -1,0 +1,208 @@
+package cyral
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/cyralinc/terraform-provider-cyral/client"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var allAuthSchemes = []string{
+	"environment_variable",
+}
+
+type AuthSchemeEnvironmentVariable struct {
+	VariableName string `json:"variableName"`
+}
+
+type AuthScheme struct {
+	EnvironmentVariable *AuthSchemeEnvironmentVariable `json:"environmentVariable"`
+}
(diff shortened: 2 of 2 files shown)


cyral/provider.go
```diff
@@ -170,6 +170,7 @@ func Provider() *schema.Provider {
 			"cyral_repository_datamap":          resourceRepositoryDatamap(),
 			"cyral_repository_identity_map":     resourceRepositoryIdentityMap(""),
 			"cyral_repository_local_account":    resourceRepositoryLocalAccount(),
+			"cyral_repository_user_account":     resourceRepositoryUserAccount(),
```
ricardorey10, line 173: I think we should call this `database_account` because that's the terminology we're using in Jeeves.

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,414 @@
...
+	UserAccountID string `json:"userAccountID"`
+}
+
+func (resp *CreateUserAccountResponse) WriteToSchema(d *schema.ResourceData) error {
+	d.SetId(resp.UserAccountID)
+	return nil
+}
+
+func (response *UserAccountResource) WriteToSchema(d *schema.ResourceData) error {
+	d.SetId(response.UserAccountID)
```
ricardorey10, line 90: I believe that this function will need to write all the details of the `UserAccountResource` you just read into the `d` map thingy. If you don't and you change something about this user account in the UI, then Terraform won't know about it and a subsequent `terraform apply` will wrongly think that nothing changed, and thus things will diverge.

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,414 @@
...
+		return fmt.Errorf("exactly one auth_scheme attribute is required")
+	}
+
+	authSchemes := authSchemeSet[0].(map[string]interface{})
+
+	for k, v := range authSchemes {
+		authSchemeDetails := v.(*schema.Set).List()
+		if len(authSchemeDetails) == 0 {
+			continue
+		}
```
ricardorey10, line 110: I feel like this should also return an error, but I honestly don't know what all these casts are doing. Let's see what Victor and Alex say.

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,414 @@
...
+								},
+							},
+						},
+						"hashicorp_vault": {
+							Description: "Credential option to set the repository user account from " +
+								"Hashicorp Vault.",
+							Optional: true,
+							Type:     schema.TypeSet,
+							MaxItems: 1,
+							MinItems: 1,
```
ricardorey10, line 349: Is this actually correct here even though for Vault we have 2 fields ("path" and "is_dynamic_user_account") ?

ricardorey10 (commented):
Some early comments

PR #288, opened 2022-09-19
ENG-9385: User Account resource for Gatekeeper Project

hbaackmann:
## Description of the change

> This PR adds a new resource definition for user accounts. The api definitions for user accounts can be found [here](https://github.com/cyralinc/gatekeeper/blob/deb614c9ab45b4d63f13659e1f9df7f294389ec5/proto/gatekeeper.proto#L27-L80).  No tests have been added yet.

## Type of change

### Development

- [x] Lint rules pass locally
- [] The code changed/added as part of this pull request has been covered with tests
- [] All tests related to the changed code pass in development

### Code review

### Testing

> Description of testing


cyral/provider.go
```diff
@@ -171,6 +170,7 @@ func Provider() *schema.Provider {
 			"cyral_repository_datamap":               resourceRepositoryDatamap(),
 			"cyral_repository_identity_map":          resourceRepositoryIdentityMap(""),
 			"cyral_repository_local_account":         resourceRepositoryLocalAccount(),
+			"cyral_repository_user_account":          resourceRepositoryUserAccount(),
```
ricardorey10, line 173: I think we should deal with "database accounts" instead of "user accounts" (at least in user-facing stuff), because that's what we're calling user accounts in the UI.

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+				},
+			})
+			if err != nil {
+				return fmt.Errorf("error setting 'approval_config': %w", err)
+			}
+		}
+	}
+
+	var authScheme []interface{}
+	if resource.AuthScheme.AWSIAM != nil {
```
ricardorey10, line 115: Might be cleaner to do this as a switch:

```
switch {
case resource.AuthScheme.AWSIAM != nil:
...
}
```

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+		)
+	}
+
+	authSchemes := authSchemeSet[0].(map[string]interface{})
+
+	for k, v := range authSchemes {
+		authSchemeDetails := v.(*schema.Set).List()
+		if len(authSchemeDetails) == 0 {
+			continue
+		}
```
ricardorey10, line 238: Can this ever happen if the auth scheme is mandatory?

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+
+			"repository_id": {
+				Description: "ID of the repository.",
+				Required:    true,
+				Type:        schema.TypeString,
+				ForceNew:    true,
+			},
+
+			"name": {
+				Description: "The name of the User Account.",
```
ricardorey10, line 414: User Account -> Database Account

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+				Required:    true,
+				Type:        schema.TypeString,
+			},
+
+			"auth_database_name": {
+				Description: "For User Accounts associated with a MongoDB cyral_repository, " +
+					"This value is only meaningful for User Accounts associated with " +
```
ricardorey10, line 421: The "T" in "This" should be lowercase. That said, this description doesn't quite explain what this is for. Maybe you can ask Pramod exactly how this is used (I think in Mongo you can have multiple user accounts with the same name but, as long as the auth_database_name is different, they count as different accounts or something like that... but Pramod can give you a better explanation).

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+
+			"approval_config": {
+				Description: "Configurations related to Approvals.",
+				Optional:    true,
+				Type:        schema.TypeSet,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"automatic_grant": {
+							Description: "If true, it is possible to create an approval with an automatic grant" +
```
ricardorey10, line 435: I think you're missing a white space after "grant".

Also, maybe this can be rephrased like "If true, it is possible for approvals to be automatically granted" or something like that?

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+								"associated with this User Account.",
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"max_auto_grant_duration": {
+							Description: "The maximum duration in seconds for the lifespan of the approval, " +
+								"if automatic_grant is true. E.g.: '2000s', '3000.5s'",
```
ricardorey10, line 442: Mhm, this makes it sound like you cannot make a request for an approval with a longer duration than this variable.

What do you think about

> The maximum duration in seconds for which approvals can be automatically granted

?

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+		}
+	}
+
+	// Handle Auth Scheme (required field).
+	authSchemeSet := d.Get("auth_scheme").(*schema.Set).List()
+	if len(authSchemeSet) != 1 {
+		return fmt.Errorf(
+			"exactly one auth_scheme attribute is required",
+		)
+	}
```
VictorGFM, line 230: This is already validated by the Required: true in the `auth_scheme`.
```suggestion
```

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+		authSchemeDetails := v.(*schema.Set).List()
+		if len(authSchemeDetails) == 0 {
+			continue
+		}
+		if len(authSchemeDetails) > 1 {
+			return fmt.Errorf(
+				"exactly one of %v is required in auth_scheme",
+				allAuthSchemes,
+			)
+		}
```
VictorGFM, line 244: This can be replaced by the use of the `ExactlyOneOf` in the resource schema.
```suggestion
```

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+					Password: m["password"].(string),
+				},
+			}
+		case "hashicorp_vault":
+			userAccount.AuthScheme = &AuthScheme{
+				HashicorpVault: &AuthSchemeHashicorpVault{
+					Path: m["path"].(string),
+				},
+			}
+			if m["is_dynamic_user_account"] != nil {
```
VictorGFM, line 278: `is_dynamic_user_account` should probably be a required field.

VictorGFM (commented):
This looks really good @hbaackmann! I provided a few comments that I believe will help improve the resource. I think the tests will come later on right? Could you also add the docs for the resource ([see](https://github.com/cyralinc/terraform-provider-cyral#updating-the-documentation)), thanks.

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+				MinItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"environment_variable": {
+							Description: "Credential option to set the repository user account from " +
+								"Environment Variable.",
+							Optional: true,
+							Type:     schema.TypeSet,
+							MaxItems: 1,
+							MinItems: 1,
```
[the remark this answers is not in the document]
Yowgf: @VictorGFM I don't think it is possible to use `ExactlyOneOf` in a schema nested inside TypeSet. I have struggled with this in the past and found no way around it. See [this comment](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/schema.go#L278-L283)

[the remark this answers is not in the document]
Yowgf: If we use TypeList with `MaxItems: 1` for `auth_scheme`, then it is possible. I think this would make sense. There is no difference between TypeList with `MaxItems: 1` and TypeSet with `MaxItems: 1` from the user's perspective, so it should be OK. LMK your thoughts.

cyral/provider.go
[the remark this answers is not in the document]
Yowgf: I agree with @VictorGFM here.

Yowgf (commented):
I took a quick look @hbaackmann and mainly added some comments under Victor's comments. Regarding the resource name, I agree with Victor and the way it is now: just leave it as `cyral_repository_user_account`.

cyral/provider.go
hbaackmann: Ah, I see!

cyral/resource_cyral_repository_user_account.go
hbaackmann: Yes, basically `authSchemeSet[0].(map[string]interface{})` always returns a map with length 7, containing where keys to all possible `auth_scheme` types are present, and each key corresponds to a `*schema.Set`. However, the scheme.Set has a length of 0 for all `auth_scheme` types except for the one that was populated.

cyral/resource_cyral_repository_user_account.go
hbaackmann: Oops, forgot to finish editing that line! Thanks :)

cyral/provider.go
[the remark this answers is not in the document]
wcmjunior: I think we should stick to the API, in this sense `cyral_repository_user_account` is more appropriate. Please make sure our public docs related to `Database Accounts` will refer to this resource. Also, in the resource documentation in the provider itself, just add an observation that this relates to `database accounts` in order to allow those users that landed on the provider to locate themselves on what this resource looks like in the UI.

hbaackmann: I think I am going to keep it matching the API. From @wcmjunior:
`Historically our UI has some influences that go well beyond the engineering team, meaning that it can change over time. If something is called Database account today it could be changed to something else in the future due to some UI change for some specific new workflow/feature/other requirement and the API could remain untouched.`

cyral/resource_cyral_repository_user_account.go
hbaackmann: thanks!

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+				"kubernetes_secret": []interface{}{
+					map[string]interface{}{
+						"secret_name": resource.AuthScheme.KubernetesSecret.SecretName,
+						"secret_key":  resource.AuthScheme.KubernetesSecret.SecretKey,
+					},
+				},
+			},
+		}
+	} else {
+		return fmt.Errorf("auth scheme is required, user account is corrupt: %v", resource)
```
wcmjunior, line 188: This is already a `required` parameter. This code should never be reachable.
```suggestion
```

cyral/resource_cyral_repository_user_account.go
wcmjunior: ...and this `if` should be removed. :)

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+			},
+
+			"name": {
+				Description: "The name of the User Account.",
+				Required:    true,
+				Type:        schema.TypeString,
+			},
+
+			"auth_database_name": {
+				Description: "For User Accounts associated with a MongoDB cyral_repository, " +
```
wcmjunior, line 420:
```suggestion
				Description: "For User Accounts associated with a MongoDB `cyral_repository`, " +
```

The descriptions can use markdown notation. We usually highlight all the provider resources, links and other stuff in order to make the documentation easier to follow. References to our public docs are also welcome where they make sense. :-)

cyral/resource_cyral_repository_user_account.go
wcmjunior, line 435:
```suggestion
							Description: "If `true`, it is possible to create an approval with an automatic grant" +
```

Same cosmetic markdown suggestion as before. :)

cyral/resource_cyral_repository_user_account.go
wcmjunior, line 442:
```suggestion
								"if automatic_grant is true. E.g.: `\"2000s\"`, `\"3000.5s\"",
```

wcmjunior (commented):
Not much to comment regarding the code as all the previous reviews are thorough, but some minor observations on usage of markdown in the scheme `Description` in order to keep the automatically-created docs easier to read.

cyral/provider.go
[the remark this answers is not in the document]
hbaackmann: Awesome, I will keep it as User Accounts. Wilson also agrees, for the following reason: 

> Historically our UI has some influences that go well beyond the engineering team, meaning that it can change over time. If something is called Database account today it could be changed to something else in the future due to some UI change for some specific new workflow/feature/other requirement and the API could remain untouched.

cyral/resource_cyral_repository_user_account.go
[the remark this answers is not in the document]
hbaackmann: Yes, I was also running into issues with `ExactlyOneOf` in a TypeSet. Even though TypeList is generally supposed to be used for ordered lists, as @Yowgf mentioned, this doesn't matter if there can only be one element. I will try that out.

cyral/resource_cyral_repository_user_account.go
```diff
@@ -0,0 +1,596 @@
...
+						"automatic_grant": {
+							Description: "If true, it is possible to create an approval with an automatic grant" +
+								"associated with this User Account.",
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"max_auto_grant_duration": {
+							Description: "The maximum duration in seconds for the lifespan of the approval, " +
+								"if automatic_grant is true. E.g.: '2000s', '3000.5s'",
+							Optional: true,
```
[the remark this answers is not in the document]
hbaackmann: Ahh I see, validation is done in the backend-- I just looked at proto definition. Great catch!!!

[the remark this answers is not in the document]
hbaackmann: Ahh, just found my answer. Will change this.

cyral/resource_cyral_repository_user_account.go
hbaackmann: Yes, I added a bunch of error catching while developing to make sure I could figure out what is going on. Do you think I should remove these cases, since we know that there should be no error? Just to make it clear when reading the code what is expected.

VictorGFM: Personally, Im usually in favor of removing this type of validation if its already being done in the schema since I think defining the validation in the schema is more "terraform idiomatic". But this is just an opinion that I have, both options will work in the end so feel free to decide about that

wcmjunior (approved):

PR #303, opened 2022-10-13
Cyral Terraform 3.0 Migration

hbaackmann:
## Description of the change

> This PR introduces a guide and script needed to upgrade to Cyral Terraform Provider 3.0.0.

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review

### Testing

> Description of testing


scripts/3.0-migration.sh
```diff
@@ -0,0 +1,252 @@
...
+done
+for ((i = 0; i < ${#access_rule_resource_names[@]}; i++));do
+    terraform state show -no-color ${access_rule_resource_names[$i]} | grep -v "   id " >> cyral_migration_repository_access_rules_and_user_accounts.txt
+done
+
+mv cyral_migration_repository_access_rules_and_user_accounts.txt cyral_migration_repository_access_rules_and_user_accounts.tf
+terraform fmt
+
+echo; echo; echo;
+echo "Now that the Terraform state is up-to-date, lets clean up your .tf file."
```
ricardorey10, line 189: lets -> let's

ricardorey10 (commented):
Just some comments about the guide itself - I'm not brave enough to go through the bash code :stuck_out_tongue_closed_eyes:  -.

If you could make a recording of this whole process that would be amazing. It would really clarify how things work - i.e when users need to stop, move to a different terminal, run some other command, go back to the script, etc -, and might even be useful for training the folks who will assist our customers with this migration.

scripts/3.0-migration.sh
```diff
@@ -0,0 +1,252 @@
...
+read -p "Are you ready to revert to your pre-migration Terraform state? [N/y] " -n 1 -r
+echo    # move to a new line
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    echo "Exiting..."
+    [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
+fi
+
+# Downgrade Terraform
+terraform init -upgrade
```
VictorGFM, line 246: This would only downgrade the terraform provider version if users change back the version constraint on the .tf file, so we might want to add that instruction before running this command.

VictorGFM (commented):
Hey @hbaackmann I've addressed Ricardo's comment and a few small details that I notice in the doc and the script. I also added a comment below regarding the terraform provider downgrade. Aside from that, everything looks great! 
I still would like to do some tests for the script before approving the PR, as I didn't get the chance to run it yet on a real scenario, but I would say that I dont expect to spot any issues since from what I reviewed things look good.

scripts/3.0-migration.sh
VictorGFM: Oh, I just noticed that the backup file for the .tf config will have the old version and when users replace the .tf with the backup content the downgrade will always work, please ignore the comment above :)

VictorGFM:
This PR also includes an update in the cyral_repository_access_rules resource, to not allow empty policy_ids to be set on the terraform state during terraform import. This was needed due to a terraform issue when using nested empty TypeLists. For reference, see the following discussion:
- https://cyralinc.slack.com/archives/C01NQPCDRTQ/p1666040013105129

VictorGFM (approved):
LGTM 🙂

PR #361, opened 2023-02-18
Improve cyral_repository docs

wcmjunior:
## Description of the change

Improve `cyral_repository` docs improving `dynamic` and `mongodb_settings` attributes.

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review

### Testing

NA

### DIFF
--- a/cyral/resource_cyral_repository.go
+++ b/cyral/resource_cyral_repository.go
@@ -442,21 +442,25 @@ func resourceRepository() *schema.Resource {
 						RepoMongoDBServerTypeKey: {
 							Description: "Type of the MongoDB server. Allowed values: " + supportedTypesMarkdown(mongoServerTypes()) +
 								".\nThe following conditions apply:\n" +
-								"  - If `" + RepoMongoDBServerTypeKey + "=" + Standalone + "`, then only one `" + RepoNodesKey +
-								"` block can be declared and it must be static. See `" + RepoNodeDynamicKey + "`." +
-								"  - If `" + RepoMongoDBServerTypeKey + "=" + Sharded + "` and the a `" + RepoMongoDBSRVRecordName +
-								"` is provided, then all `" + RepoNodesKey + "` blocks must be declared dynamic, " +
-								"otherwise all of them must be static. See `" + RepoNodeDynamicKey + "`.",
+								"  - If `" + Sharded + "` and `" + RepoMongoDBSRVRecordName + "` *not* provided, then all `" +
+								RepoNodesKey + "` blocks must be static (see `" + RepoNodeDynamicKey + "`).\n" +
+								"  - If `" + Sharded + "` and `" + RepoMongoDBSRVRecordName + "` provided, then all `" +
+								RepoNodesKey + "` blocks must be dynamic (see `" + RepoNodeDynamicKey + "`).\n" +
+								"  - If `" + Standalone + "`, then only one `" + RepoNodesKey +
+								"` block can be declared and it must be static (see `" + RepoNodeDynamicKey + "`). The `" +
+								RepoMongoDBSRVRecordName + "`is not supported in this configuration.\n" +
+								"  - If `" + ReplicaSet + "` and `" + RepoMongoDBSRVRecordName + "` *not* provided, then `" +
+								RepoNodesKey + "` blocks may mix dynamic and static nodes (see `" + RepoNodeDynamicKey + "`).\n" +
+								"  - If `" + ReplicaSet + "` and `" + RepoMongoDBSRVRecordName + "` provided, then `" +
+								RepoNodesKey + "` blocks must be dynamic (see `" + RepoNodeDynamicKey + "`).\n",
 							Type:         schema.TypeString,
 							Required:     true,
 							ValidateFunc: validation.StringInSlice(mongoServerTypes(), false),
 						},
 						RepoMongoDBSRVRecordName: {
-							Description: "Name of a DNS SRV record which contains cluster topology details. The following conditions apply:\n" +
-								"  - If `" + RepoMongoDBSRVRecordName + "` is specified, then all `" + RepoNodesKey + "` blocks can be declared `" +
-								RepoNodeDynamicKey + "=true`.\n" +
-								"  - If `" + RepoMongoDBSRVRecordName + "` is *not* specified, then at *least one* node must have `" + RepoNodesKey +
-								"." + RepoHostKey + "` and `" + RepoNodesKey + "." + RepoPortKey + "`.\n",
+							Description: "Name of a DNS SRV record which contains cluster topology details. " +
+								"If specified, then all `" + RepoNodesKey + "` blocks must be declared dynamic " +
+								"(see `" + RepoNodeDynamicKey + "`).",
 							Type:     schema.TypeString,
 							Optional: true,
 						},
--- a/cyral/resource_cyral_repository.go
+++ b/cyral/resource_cyral_repository.go
@@ -409,9 +409,20 @@ func resourceRepository() *schema.Resource {
 							Optional:    true,
 						},
 						RepoNodeDynamicKey: {
-							Description: "Indicates if node is dynamically discovered. If true, `host` and `port` must be empty.",
-							Type:        schema.TypeBool,
-							Optional:    true,
+							Description: "*Only supported for MongoDB in cluster configurations.*\n" +
+								"Indicates if the node is dynamically discovered, meaning that the sidecar " +
+								"will query the cluster to get the topology information and discover the " +
+								"hosts of the remaining nodes. If set to `true`, `host` and `port` must " +
+								"be empty. A node that does not declare this field is considered `static`.\n" +
+								"The feature works in the following conditions: \n" +
+								"  - The total number of declared `" + RepoNodesKey + "` blocks must match " +
+								"the actual number of nodes in the cluster.\n" +
+								"  - If there are static nodes in the configuration, they must be declared " +
+								"before all dynamic nodes.\n" +
+								"  - See the MongoDB-specific configuration in the [" + RepoMongoDBSettingsKey +
+								"](#nested-schema-for-" + RepoMongoDBSettingsKey + ").",
+							Type:     schema.TypeBool,
+							Optional: true,
 						},
 					},
 				},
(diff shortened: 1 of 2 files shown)


docs/resources/repository.md
```diff
@@ -141,9 +146,13 @@ Required:
 - `server_type` (String) Type of the MongoDB server. Allowed values:
   - `replicaset`
   - `standalone`
-  - `sharded`
+  - `sharded`.
+    The following conditions apply:
+  - If `server_type=standalone`, then only one `repo_node` block can be declared and it must be static. See `dynamic`. - If `server_type=sharded` and the a `srv_record_name` is provided, then all `repo_node` blocks must be declared dynamic, otherwise all of them must be static. See `dynamic`.
```
ricardorey10, line 151: The indentation of the second `- If` looks different than the first one, but this might just be my browser. Please check before merging.

docs/resources/repository.md
```diff
@@ -141,9 +146,13 @@ Required:
...
 
 Optional:
 
 - `replica_set_name` (String) Name of the replica set, if applicable.
-- `srv_record_name` (String) Name of a DNS SRV record which contains cluster topology details
+- `srv_record_name` (String) Name of a DNS SRV record which contains cluster topology details. The following conditions apply:
+  - If `srv_record_name` is specified, then all `repo_node` blocks can be declared `dynamic=true`.
+  - If `srv_record_name` is _not_ specified, then at _least one_ node must have `repo_node.host` and `repo_node.port`.
```
ricardorey10, line 158: I feel like here we're kinda giving them mixed instructions. Here we say:

> If `srv_record_name` is _not_ specified, then at _least one_ node must have `repo_node.host` and `repo_node.port`.

This means that I can mix static and dynamic, right? However, above, we have:

> If `server_type=sharded` and the a `srv_record_name` is provided, then all `repo_node` blocks must be declared dynamic, otherwise all of them must be static. See `dynamic`

This means that we *cannot* mix.

So, if people read the former, they might be tempted to mix stuff (which is wrong, and which will result in the API returning an error), and if they read the latter, they will do the right thing. If they read both... then they will be confused :stuck_out_tongue_winking_eye: 

Maybe we should repeat the part about not mixing static and dynamic in the former?

cyral/resource_cyral_repository.go
```diff
@@ -409,9 +409,20 @@ func resourceRepository() *schema.Resource {
...
+								"hosts of the remaining nodes. If set to `true`, `host` and `port` must " +
+								"be empty. A node that does not declare this field is considered `static`.\n" +
+								"The following conditions apply: \n" +
+								"  - The total number of declared `" + RepoNodesKey + "` blocks must match " +
+								"the actual number of nodes in the cluster.\n" +
+								"  - If there are static nodes in the configuration, they must be declared " +
+								"before all dynamic nodes.\n" +
+								"  - See the MongoDB-specific configuration in the [" + RepoMongoDBSettingsKey +
+								"](#nested-schema-for-" + RepoMongoDBSettingsKey + ").",
+							Type:     schema.TypeBool,
```
deepak-cyral, line 424: Is it possible (and desirable) to specify default value (since the parameter is optional)?

cyral/resource_cyral_repository.go
```diff
@@ -409,9 +409,20 @@ func resourceRepository() *schema.Resource {
 							Optional:    true,
 						},
 						RepoNodeDynamicKey: {
-							Description: "Indicates if node is dynamically discovered. If true, `host` and `port` must be empty.",
-							Type:        schema.TypeBool,
-							Optional:    true,
+							Description: "*Only supported for MongoDB in cluster configurations.*\n" +
+								"Indicates if the node is dynamically discovered, meaning that the sidecar " +
+								"will query the cluster to get the topology information and discover the " +
+								"hosts of the remaining nodes. If set to `true`, `host` and `port` must " +
```
deepak-cyral, line 415:
```suggestion
								"addresses of the dynamic nodes. If set to `true`, `host` and `port` must " +
```

cyral/resource_cyral_repository.go
```diff
@@ -429,15 +440,25 @@ func resourceRepository() *schema.Resource {
...
 							Type:         schema.TypeString,
 							Required:     true,
 							ValidateFunc: validation.StringInSlice(mongoServerTypes(), false),
 						},
 						RepoMongoDBSRVRecordName: {
-							Description: "Name of a DNS SRV record which contains cluster topology details",
-							Type:        schema.TypeString,
-							Optional:    true,
+							Description: "Name of a DNS SRV record which contains cluster topology details. The following conditions apply:\n" +
+								"  - If `" + RepoMongoDBSRVRecordName + "` is specified, then all `" + RepoNodesKey + "` blocks can be declared `" +
```
deepak-cyral, line 456:
```suggestion
								"  - If `" + RepoMongoDBSRVRecordName + "` is specified, then all `" + RepoNodesKey + "` blocks must be declared `" +
```

wcmjunior: @deepak-cyral , per our previous conversation I understood that all nodes must only be dynamic for `sharded` types. Do you also mean that `replicaset` types must also use `dynamic=true` if the SRV record is provided?

docs/resources/repository.md
wcmjunior: You are right. It is missing a new line.

docs/resources/repository.md
wcmjunior: > This means that I can mix static and dynamic, right?
Yes, this is what I understood from this internal statement: `at least one node must be static (not dynamic) UNLESS SRV record is specified in mongodb settings`, which means that we should be able to mix dynamic and static if the SRV record is not present and as long as the first node is static.

Now I am looking at [this code](https://github.com/cyralinc/crud/blob/main/repo/repo_crud.go#L125-L156), I see that if the `srv_record_name` is **not** specified, we check:
- If cluster type is `sharded`, then all nodes must be static;
- If cluster type is `standalone`,  then a single node must be declared and it needs to be static (following [this check](https://github.com/cyralinc/crud/blob/main/repo/repo_crud.go#L155) and [this one](https://github.com/cyralinc/crud/blob/22ba2cb912d6f8cee98d2970e586d558776587f6/repo/repo_crud.go#L1549)).
- As `replicaset` is not ruled out from that code, [this line](https://github.com/cyralinc/crud/blob/main/repo/repo_crud.go#L155) allows us to have mixed configurations of static+dynamic as long as the first node is static.

By contrast, if `srv_record_name` is specified, we check [here](https://github.com/cyralinc/crud/blob/22ba2cb912d6f8cee98d2970e586d558776587f6/repo/repo_crud.go#L158-L167):
- All nodes must be dynamic.


I will update the docs accordingly to those rules. LMK if I missed something.

wcmjunior: Given the rules above, we could simplify the documentation and move it to the proper attributes as follows:

`srv_record_name`
- If provided, then all nodes must be dynamic .

`server_type`
- If `sharded` and `srv_record_name` not provided, then all nodes must be static.
- If `standalone` and `srv_record_name` not provided, then only a single static node must be declared.
- If `replicaset` and `srv_record_name` not provided, then mixed configuration of static and dynamic nodes are allowed as long as the first node is static.

cyral/resource_cyral_repository.go
wcmjunior: AFAIK this `TypeBool` is mapped to an underlying `bool` not a `*bool`, so it should not allow nulls. In this sense, I understand that the default here should be `false` then. If that is the case, then this should be the proper value for the default correct? I will double check this, though.

cyral/resource_cyral_repository.go
wcmjunior: Please, see the following conversation: https://github.com/cyralinc/terraform-provider-cyral/pull/361#discussion_r1110482324

cyral/resource_cyral_repository.go
wcmjunior: Having second thoughts on my previous argument and I see now it does not make sense from the user standpoint. We should indeed provide the default values in the code as they would reflect in the docs and give a better notion to our users what is the behaviour when a given argument is not provided. Anyway, I think we should address these defaults in a different PR as this would require some tests. @ricardorey10 , could you take a stab on providing these default values to the schema?

cyral/resource_cyral_repository.go
```diff
@@ -409,9 +409,20 @@ func resourceRepository() *schema.Resource {
...
 						},
 						RepoNodeDynamicKey: {
-							Description: "Indicates if node is dynamically discovered. If true, `host` and `port` must be empty.",
-							Type:        schema.TypeBool,
-							Optional:    true,
+							Description: "*Only supported for MongoDB in cluster configurations.*\n" +
+								"Indicates if the node is dynamically discovered, meaning that the sidecar " +
+								"will query the cluster to get the topology information and discover the " +
+								"addresses of the dynamic nodes. If set to `true`, `host` and `port` must " +
+								"be empty. A node that does not declare this field is considered `static`.\n" +
```
deepak-cyral, line 416:
```suggestion
								"be empty. A node with value of this field as false considered `static`.\n" +
```

cyral/resource_cyral_repository.go
```diff
@@ -409,9 +409,20 @@ func resourceRepository() *schema.Resource {
...
+								"The following conditions apply: \n" +
+								"  - The total number of declared `" + RepoNodesKey + "` blocks must match " +
```
deepak-cyral, line 418: This particular point probably belongs as description for `RepoNodesKey`

deepak-cyral (approved):
Approving so we don't hold this up - please make changes (and accept/reject suggestions) as you see fit.

cyral/resource_cyral_repository.go
ricardorey10: Sure thing. I'll put my thoughts here for discussion, and we can create a PR later (I'm a bit tied up with some Snowflake stuff now so I can't volunteer to do that change right now, but if it can wait a few days I can take it):

- "host": this should default to an empty string, as we cannot guess anything better.
- "port": this should default to zero, as we cannot guess anything.
- "dynamic": I think the default should be `false`. The reasoning here is that, if I supply the host and port but I don't supply `dynamic`, then this will result in an API error, because we can't have `dynamic = true` and non-empty host:port. This will also make it obvious that, for dynamic repo nodes, the *only* thing they have to configure is the dynamic-ness os the node (as `true`).

With the above defaults, we can have the following configurations:

```
repo_node {
  dynamic = true   // Host and port are empty, which is fine
}
```

```
repo_node {
  host = "something"    // dynamic is false, which is what we want
  port = 1234
}
```

```
repo_node {
  host = "something"    // Just combining the two above in a single config
  port = 1234
}

repo_node {
  dynamic = true
}
```

cyral/resource_cyral_repository.go
```diff
@@ -429,15 +440,31 @@ func resourceRepository() *schema.Resource {
...
+								RepoNodesKey + "` blocks must be dynamic (see [`" + RepoNodeDynamicKey + "`](#" + RepoNodeDynamicKey + ")).\n",
 							Type:         schema.TypeString,
 							Required:     true,
 							ValidateFunc: validation.StringInSlice(mongoServerTypes(), false),
 						},
 						RepoMongoDBSRVRecordName: {
-							Description: "Name of a DNS SRV record which contains cluster topology details",
-							Type:        schema.TypeString,
-							Optional:    true,
+							Description: "Name of a DNS SRV record which contains cluster topology details. " +
```
ricardorey10, line 461: This should default to an empty string. The backend will interpret that as "no SRV record supplied" and thus it will go through the "hard-coded" path.

cyral/resource_cyral_repository.go
```diff
@@ -429,15 +440,31 @@ func resourceRepository() *schema.Resource {
 							Optional:    true,
```
ricardorey10, line 440: Here, the ideal behavior should be "if the server type is replicaset, then this is actually mandatory". If we can express that in code, that would be nice.

Regardless, the default here should just be an empty string.

ricardorey10 (approved):
@deepak-cyral @wcmjunior I left some comments about the defaults.

Regardless of those default values, though, I think this PR looks good for merging, and we can take the default stuff in a clean PR after this is merged.

PR #441, opened 2023-08-21
ENG-12435 -Uptake of new Listener API with override repo tls flag and tls mode.

gengdahlCyral:
Update terraform providers to use new Listener API with override repo tls flag and tls mode.

## Description of the change

> Description here

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review

### Testing

> Description of testing


cyral/resource_cyral_sidecar_listener.go
```diff
@@ -402,5 +418,22 @@ func getSidecarListenerSchema() map[string]*schema.Schema {
...
+				"Default value generated by API if not provided. ",
+			Type:     schema.TypeBool,
+			Optional: true,
+			Computed: true,
+		},
+		TlsModeKey: {
+			Description: "TLS mode. " +
+				"Default value generated by API if not provided. " +
+				"Note! This field is in effect only if OverrideRepoClientTlsSettings is set to true or the listener " +
+				"is a SMART port (enabled in more than one binding). " +
```
VictorGFM, line 432: **Suggestion:** Is there a link in our public doc regarding smart ports? Might be interesting to add this link in this description as a reference for customers that want to understand how it works. I would also recommend adding a reference to the `cyral_repository_binding` resource here, something like:
```suggestion
				"is a SMART port (enabled in more than one binding - see [`cyral_repository_binding`](../resources/repository_binding.md)). " +
```

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -402,5 +418,22 @@ func getSidecarListenerSchema() map[string]*schema.Schema {
...
 			},
 		},
+		OverrideRepoClientTlsSettingsKey: {
+			Description: "Override TLS settings defined in the repo. " +
+				"Default value generated by API if not provided. ",
+			Type:     schema.TypeBool,
+			Optional: true,
+			Computed: true,
+		},
+		TlsModeKey: {
```
VictorGFM, line 428: @gengdahlCyral could you please also add validation for the possible values that are valid for this field? I understand that the API already does this validation and that it is easier to maintain if we don't duplicate the validation here in the terraform provider, but since the validation messages returned by the API are often not clear enough, adding a validation that is computed during the terraform plan instead of during the execution, and that also provides a proper message that explicitly mentions the error and which values are allowed for this field would improve the terraform provider UX overall, which is something we are aiming for in a recent conversation that I had with @wcmjunior. The validation could be something like:
```Go
ValidateFunc: validation.StringInSlice(tlsModes(), false),
```

cyral/resource_cyral_sidecar_listener.go
```diff
@@ -402,5 +418,22 @@ func getSidecarListenerSchema() map[string]*schema.Schema {
...
+			Description: "TLS mode. " +
+				"Default value generated by API if not provided. " +
+				"Note! This field is in effect only if OverrideRepoClientTlsSettings is set to true or the listener " +
```
VictorGFM, line 431:
```suggestion
				"Note! This field is in effect only if `override_repo_client_tls_settings` is set to `true` or the listener " +
```

VictorGFM (commented):
Changes look good @gengdahlCyral, I just have a few suggestions to improve the resource usability and the docs a little bit.

cyral/resource_cyral_sidecar_listener.go
wcmjunior: Regarding the link to our public docs, please refer to the latest version instead of some specific release.

PR #450, opened 2023-09-08
ENG-12558: add AuthType to the repo conf auth payload

ricardorey10:
## Description of the change

The `conf/auth` payload got a new field, `AuthType`, so this PR adds it so that it is configurable via TF.

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review

### Testing

Tested in a k8s deployment + unit tests


cyral/resource_cyral_repository_conf_auth.go
```diff
@@ -204,6 +210,13 @@ func repositoryConfAuthResourceSchemaV0() *schema.Resource {
 				Optional:    true,
 				Default:     defaultRepoTLS,
 			},
+			"auth_type": {
+				Description: "Authentication type for this repository. 0 for Opaque Token, " +
+					"1 for AWS IAM. Defaults to 0",
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  defaultAuthType,
+			},
```
VictorGFM, line 219: Lets also add a `ValidateFunc` here to ensure that the enum validation is also done in the terraform provider. I understand that the API already validates that, but doing that directly in the terraform provider enhances the UX by validating during terraform plan instead of during execution, and also gives us flexibility to provide better error messages to users. Adding these kinds of validations in the provider is something that we opted to do in favor of a better UX for customers.
```suggestion
			"auth_type": {
				Description: "Authentication type for this repository. 0 for Opaque Token, " +
					"1 for AWS IAM. Defaults to 0",
				Type:     schema.TypeInt,
				Optional: true,
				Default:  defaultAuthType,
				ValidateFunc: validation.StringInSlice(repoAuthTypes(), false),
			},
```

cyral/resource_cyral_repository_conf_auth.go
```diff
@@ -204,6 +210,13 @@ func repositoryConfAuthResourceSchemaV0() *schema.Resource {
 				Optional:    true,
 				Default:     defaultRepoTLS,
 			},
+			"auth_type": {
+				Description: "Authentication type for this repository. 0 for Opaque Token, " +
+					"1 for AWS IAM. Defaults to 0",
+				Type:     schema.TypeInt,
```
VictorGFM, line 216: For enum fields, we usually define them as `schema.TypeString`. I believe this makes the valid values more explicit and self-explanatory in some sense. We could do the same here and also use the `supportedTypesMarkdown` function in the field description to explicitly list the valid values for this field.

docs/index.md
```diff
@@ -80,12 +80,9 @@ the old one.
...
 
-### Required
-
-- `control_plane` (String) Control plane host and API port (ex: `tenant.app.cyral.com`)
-
 ### Optional
 
 - `client_id` (String, Sensitive) Client id used to authenticate against the control plane. Can be ommited and declared using the environment variable `CYRAL_TF_CLIENT_ID`.
 - `client_secret` (String, Sensitive) Client secret used to authenticate against the control plane. Can be ommited and declared using the environment variable `CYRAL_TF_CLIENT_SECRET`.
+- `control_plane` (String) Control plane host and API port (ex: `tenant.app.cyral.com`)
```
VictorGFM, line 87: I believe this change was auto-generated, which is weird because it is incorrect. The `control_plane` is actually defined as a `required` field in the provider schema, in the `provider.go` file. My guess is that this could be happening due to the `DefaultFunc` used for this field in order for it to work with env vars as well, but Im not sure. Is also strange the fact that this doesn't seem to happen when running the tfplugindocs using docker.

docs/data-sources/datalabel.md
```diff
@@ -1,6 +1,6 @@
 ---
 # generated by https://github.com/hashicorp/terraform-plugin-docs
-page_title: "cyral_datalabel Data Source - cyral"
+page_title: "cyral_datalabel Data Source - terraform-provider-cyral"
```
VictorGFM, line 3: Could you please also update the docker-compose.yaml file to use terraform-provider-cyral as the root for the project, something like:
```docker
  app:
    image: golang:1.19
    container_name: terraform_provider_cyral
    volumes:
      - .:/go/src/terraform-provider-cyral
    working_dir: /go/src/terraform-provider-cyral
```
This will make sure that no unnecessary diffs are generated in the future, regardless if one is running `tfplugindocs` from docker or locally.

VictorGFM (commented):
Thanks for working on these changes @ricardorey10, I have a few comments for your consideration below.

docs/index.md
VictorGFM: I actually tried to run `tfplugindocs generate` locally (`0.16.0`) and it actually generated the index.md correctly (with the `control_plane` as a required field), so Im not exactly sure what caused this change here.

VictorGFM (approved):
Changes look good, thanks @ricardorey10!

ricardorey10:
@VictorGFM thanks for the approval! Can you merge it, please? I don't have merge privileges for this repo :stuck_out_tongue:

